PR #64, opened 2015-04-28
Fix typos and small issues after proofreading units 0, 1 and 2.

fulv:


pbauer:
You should make all changes to the branch plone5
Am 28.04.2015 06:41 schrieb "Fulvio Casali" <notifications@github.com>:

fulv:
What is the master branch for then?

pbauer:
The master-branch is the rendered version at http://plone-training.readthedocs.org and only covers Plone 4. The plone5 branch is not yet rendered anywhere and only covers Plone 5. We should keep both branches rendered somehow but I don't jet know how.

svx:
I am still in for include them into docs.plone.org/training or training.docs.plone.org, we could just use mr.publisher to get them, build them and deploy them, in the long run I would connect them to jenkins first, we have now the first open ticket for the docs and jenkin tests ...

gforcada:
@svx what about training.plone.org that would give it a more prominent domain. As for the jenkins job... I will try to get around it soon!

svx:
@gforcada training.plone.org is fine with me ! 
Would it be handy to have a landing page, explaining about training and how to get in touch and how to give one yourself or sign up to one and such stuff ? @pbauer and @fulv what do you think ?

tkimnguyen:
Would be nice to have a reference on there about training providers & links to plone.com/providers or the equivalent on plone.org (for the time being) :)

fulv:
@svx This page contains everything you mentioned:  http://plone-training.readthedocs.org/en/latest/about.html

However, perhaps a top-level landing page could be the solution the the plone4 / plone5 problem:  it could have general information, and then split towards the two versions.

@pbauer thanks for the merge, I will repeat the same fixes on the plone5 branch.

@tkimnguyen yes, we should add the provider link on the landing page.

svx:
OK so training.plone.org -> will be a landing page with about and susbites: training for plone 4 and another one for traing plone 5 is that ok ? we will end up with something like training.plone.org/plone4 and trainig.plone.org/plone5 ?
If you want other names please say so :) !

pbauer:
@svx That sounds great!

gforcada:
@svx yes I want other names: training.plone.org/4 and training.plone.org/5 :) remove redundant info and keep them short

svx:
OK , I am bad with names, that is why I asked :)

One other thing, would it be possible to redirect http://plone-training.readthedocs.org/ to training.plone.org, once the new sites are online, that would make it better for search engines and such, I mean I know that its possible but I am not the owner of plone-training on rtd :)

pbauer:
@svx good question. It should at least be possible to connect the project with a new repo that only holds a link to the new url. But redirects are another thing...

PR #72, opened 2015-04-30
Sven typo fix and wording 

svx:
- fix various typos
- unify words [sometimes there was a word mix]
- use same words as we use on docs.plone.org [add-on] for example


pbauer:
@svx could you make the same changes in the branch plone5?

PR #80, opened 2015-05-04
Small fixes to unit 6 and 7.

fulv:
Plus, I deleted a reference to a "banner", which is either new to me, or something that no longer exists.


pbauer:
@fulv The "banner" is the element div.principal on the plone5 frontpage that says "Welcome! Plone 5 rocks!". It is droped by a diazo-rule if the item is not called 'front-page'. It is also mentioned in the next chapter.

fulv:
Thanks, I didn't connect the dots...
I reverted that commit.

PR #100, opened 2015-06-12
Small fixes for Units 31, 32 and 33

fulv:


gforcada:
Should this  be backported to master (plone 4) branch as well?

fulv:
Currently I'm working on plone 5, and I have not kept track of the differences with master, so I don't know the context of everything that changed when the 5 training got created.  I'll finish my pass through plone 5 first, and then see what to do about 4.

PR #123, opened 2015-09-18
Update zip file.

fulv:


pbauer:
done. Uses Plone 5 final now and I moved it into _static

PR #128, opened 2015-09-22
how to install/compile

thet:
this wasn't documented but isn't plain obvious.

@pbauer what's the process of contributing? all via pull requests?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,3 +4,15 @@ Mastering Plone
 Mastering Plone is intended as a week-long training for people who are new to Plone or want to learn about the current best-practices of Plone-development.
 
 Read it at http://plone-training.readthedocs.org/en/latest/index.html
+
+
+Installation
+------------
+
+To compile this training documentation to html, pdf or other formats, do this::
+
+    $ virtualenv .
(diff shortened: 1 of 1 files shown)


pbauer:
That is doumented in http://training.plone.org/5/about.html#building-the-documentation-locally

thet:
I knew, this was documented, but I didn't find it! this info is a bit hidden, did expect somethin like INSTALL.rst or being documented in README.rst.
Can be closed then...

svx:
@thet I see potential problems with this line:

To compile this training documentation to html, pdf or other formats, do this::

If I read the as user, I get the idea that I can compile the training docs into pdf, which is not the case with the current setup, in order to be able to do that, you need to change the conf.py.

See this link:

http://sphinx-doc.org/builders.html#builders

saying you have to add: rst2pdf.pdfbuilder to extensions. 

So maybe it would be better to make this clear and remove this part: , pdf or other formats, ?

pbauer:
Building the documentation locally is only really useful for trainers, everyone else should just use the website.

thet:
@sven, we can keep this closed.

PR #133, opened 2015-09-25
skip running the unifield installer. instead only unpack the buildout…

pbauer:
…-cache


smcmahon:
If it makes this any easier, there should be a file like:

http://dist.plone.org/release/###/buildout-cache.tar.bz2

where ### is the release #, for all recent and upcoming versions of Plone. It gets uploaded after the Unified Installer is built.

pbauer:
@smcmahon thanks for the tip. That is much easier indeed.

PR #144, opened 2015-10-05
Update to final buildout.  Example grepping for conflicting egg versi…

fulv:
…on dependencies.

### DIFF
--- a/buildout_1.rst
+++ b/buildout_1.rst
@@ -16,9 +16,9 @@ Topics covered:
 
 .. only:: not presentation
 
-    `Buildout <https://pypi.python.org/pypi/zc.buildout>`_ configures your application for you, according to your rules.
+    `Buildout <https://pypi.python.org/pypi/zc.buildout>`_ composes your application for you, according to your rules.
 
-    To configure your application you must define the eggs you need, which version, what configuration files Buildout has to generate for you, what to download and compile, and so on.
+    To compose your application you must define the eggs you need, which version, what configuration files Buildout has to generate for you, what to download and compile, and so on.
     Buildout downloads the eggs you requested and resolves all dependencies. You might need five different eggs, but in the end, Buildout has to install 300 eggs, all with the correct version in order to resolve all the dependencies.
 
     Buildout does this without touching your system Python or affecting any other package. The commands created by buildout bring all the required packages into the Python environment. Each command it creates may use different libraries or even different versions of the same library.
--- a/buildout_1.rst
+++ b/buildout_1.rst
@@ -16,9 +16,9 @@ Topics covered:
 
 .. only:: not presentation
 
-    `Buildout <https://pypi.python.org/pypi/zc.buildout>`_ composes your application for you, according to your rules.
+    `Buildout <https://pypi.python.org/pypi/zc.buildout>`_ configures your application for you, according to your rules.
 
-    To compose your application you must define the eggs you need, which version, what configuration files Buildout has to generate for you, what to download and compile, and so on.
+    To configure your application you must define the eggs you need, which version, what configuration files Buildout has to generate for you, what to download and compile, and so on.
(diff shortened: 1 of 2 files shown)


buildout_1.rst
```diff
@@ -342,7 +330,13 @@ Be McGuyver
 
         $ ./bin/buildout -v
 
-    If strange egg versions are requested, check the dependencies declaration of your eggs and your version pinnings.
+    If strange egg versions are requested, check the dependencies declaration of your eggs and your version pinnings.  Here is an invaluable shell command that allows you to find all packages that depend on a particular egg and version:
+
+    .. code-block:: bash
+
+        $ grep your.egg.name.here /home/vagrant/buildout-cache/eggs/*.egg/EGG-INFO/requires.txt
```
do3cc: This information is very helpful, but using an absolute path only valid for the training makes it slightly less helpful.

buildout_1.rst
```diff
@@ -16,9 +16,9 @@ Topics covered:
 
 .. only:: not presentation
 
-    `Buildout <https://pypi.python.org/pypi/zc.buildout>`_ composes your application for you, according to your rules.
+    `Buildout <https://pypi.python.org/pypi/zc.buildout>`_ configures your application for you, according to your rules.
 
-    To compose your application you must define the eggs you need, which version, what configuration files Buildout has to generate for you, what to download and compile, and so on.
+    To configure your application you must define the eggs you need, which version, what configuration files Buildout has to generate for you, what to download and compile, and so on.
     Buildout downloads the eggs you requested and resolves all dependencies. You might need five different eggs, but in the end, Buildout has to install 300 eggs, all with the correct version in order to resolve all the dependencies.
```
do3cc: IIRC I wrote this section with the wording `composes` and it catches the actions performed by buildout better than `configures`.
Quite often, buildout does not only generate configuration files but compiles nginx and so on. 
Even if it would only provide configuration, the fact that it tells plone, where to find the zeo server, where the fronted web servers find the backend and so on is more an action of composing the system than just a simple configuring an application.

buildout_1.rst
fulv: I'm not wedded to an absolute path, if you can suggest a better way.  OTOH, this is not generic documentation, it is a training.

buildout_1.rst
fulv: Yes and no:  compilation (of nginx, etc) is definitely different than configuration.  There are parts of Zope itself that need to be compiled, too.  OTOH, even the classic `configure/make/make install` cycle compiles stuff.  Perhaps we should use `install` instead of either `compose` or `configure`?
Where I disagree is that telling the system where to find zeo, front/backend, etc in my opinion is just what configuration means.  `zope.conf` is a `conf` file, not a `comp` file ;)  Also, I have never seen `compose` used in this context.

do3cc: I am not alone: http://reinout.vanrees.org/weblog/2015/06/16/djangorecipe-gunicorn.html
buildout itself does use the term assembly/assembles https://pypi.python.org/pypi/zc.buildout/2.4.4

If you tell one application where to find stuff then it is clearly configuration, but you are usually telling multiple parts where they find each other, to me this is a composition of multiple parts to work well together.
This is also the way I have always presented buildout in philips and my trainings.

fulv: Clearly we agree on the concept, just not on the word.  I'm fine reverting my change, I just won't use `compose` verbally because it does not sound right to me.  Configuration is not scope-limited to a single application;  the work we do when we set up a server stack, or even a server farm, is system configuration, not composition.

do3cc: I feel slightly bad because I could not convince you that compose is a better word. 
In the end, I cannot claim ownership for chapters I wrote and veto any changes I don't like.
The wording is important and it should be worded in such a way that the majority of the trainers feel comfortable with it and that it stays readable.
Because of that I'd love to hear opinions from @gomez @smcmahon @pbauer @macagua @fredvd @frapell @thet @giacomos

PR #145, opened 2015-10-05
Update to current bobtemplates.  package --> distribution.

fulv:


eggs1.rst
```diff
@@ -26,74 +26,70 @@ Topics covered:
...
 ..  code-block:: bash
 
-    --> Author's name [Philip Bauer]:
+    --> What kind of package would you like to create? Choose between 'Basic', 'Dexterity', and 'Theme'. [Basic]:
 
-    --> Author's email [bauer@starzel.de]:
+    --> Author's name [Fulvio Casali]:
 
-    --> Author's github username: pbauer
+    --> Author's email [fulviocasali@gmail.com]:
```
do3cc: Maybe its time to invent a generic name :-)

eggs1.rst
```diff
@@ -184,7 +176,7 @@ Before we can use our new package we have to tell Plone about it. Edit ``buildou
     test-eggs +=
         ploneconf.site [test]
 
-This tells Buildout to add the egg ``ploneconf.site``. Since it is also in the `sources` section buildout will not try to download it from pypi but will expect it in ``src/ploneconf.site``. *fs* allows you to add packages on the filesystem without a version control system, or with an unsupported one.
+This tells Buildout to add the egg ``ploneconf.site``. Since it is also in the `sources` section buildout will not try to download it from pypi but will expect it in ``src/ploneconf.site``. *fs* allows you to add eggs on the filesystem without a version control system, or with an unsupported one.
```
do3cc: I am not sure, why we need to rename packages with eggs here.
The same goes for distribution/package in the code above.
What is the general idea here, or did I miss some best practices from the official plone docs?

eggs1.rst
```diff
@@ -111,40 +107,35 @@ browser/
...
 
 ..    profiles/uninstall/
-      This folder holds another GenericSetup profile. The steps in there are executed on uninstalling.
+      This folder holds another GenericSetup profile. The steps in here are executed on uninstalling.
 
 
 .. _eggs1-include-label:
 
-Including the package in Plone
+Including the distribution in Plone
```
pbauer: this breaks rst (the underline has to be the same length)

eggs1.rst
davilima6: An egg seems to be a package with metadata and since we're talking buildout I think we'd off with the term it adopts for its keys: `eggs` and `develop-eggs`, among others.

fulv:
A Python package is a directory with an `__init__.py` file in it.  However, the second P in PyPI stands for *package*, so unfortunately in the Python world we use the term loosely.  I'm using *distribution* as defined by @optilude in his book:

> **Distribution**:  A means of packaging and distributing Python code.  A distribution consists of a directory with a `setup.py` file, Python code, and potentially other resources.  The metadata in `setup.py` may include a version, installation dependencies and licence information.
> Tools based on Setuptools/Distribute, including Buildout, can use the metadata to manage concurrent installations of different versions of a given distribution, automatically fetch dependencies, and more.
> It is common to create a distribution containing, and named after, a single package. For example, the distribution `plone.memoize` contains the package `plone.memoize` (which may of course contain sub-packages).  However, distribution names do not have to follow this pattern.

While I'm at it:

> **Egg**: A binary distribution format used by Setuptools/Distribute.  Each distribution archive is specific to a particular platform and Python version.  Therefore, it is normally safer to use a source format when distributing packages, which is just a compressed archive of the code and metadata.  (Setuptools/Distribute will turn a source distribution into an egg specific to the local development environment upon installation.)

do3cc:
I am not against this, but in that case I'd see the glossary of plone docs extended by the term distribution.
http://docs.plone.org/appendices/glossary.html#term-python-egg

fulv:
That would be good.

fulv:
I can't figure out what the test failure is about, but it can't be due to any commits in this PR.

svx:
This is because of:

(line 1339) broken    http://diazo.org - HTTP Error 502: Bad Gateway

We had these problems already yesterday something with dns between US and Europe is not working well, plone.org, dev.plone.org and someothet domains where not temp. not reachable from Europe but still from the US, also traffic from Europe was redirected via Asia and not like always straight to the US.

Nothing we can do

svx:
I have now for example also loss if I ping GitHub
mtr

pbauer:
@svx do you know why readthedocs also says that the build is broken? Something with LaTeX:
https://readthedocs.org/projects/plone-training/builds/

svx:
@pbauer there are different possible reasons I am not sure which one is responsible for that:

- could be as simple as a copy and paste error
- could be a change of utf support on the builders of rtd
- could be one or more unicode character in one of your input files that isn't mapped to an output. This may -- especially if you're using the (unicode-supporting)
- could be latex configuration in the conf.py

And this was working until 2 days ago, right ?

svx:
@fulv 

The rst syntax is not OK :) Sphinx is reporting errors:

<snip>
  training/export_code.rst:55: ERROR: Error in "code-block" directive:

  maximum 1 argument(s) allowed, 2 supplied.

  ..  code-block:: xml
</snip>
    :linenos:

fulv:
@svx:  But that syntax did not change, and it was ok before.  Unless it's complaining about something inside the code-block, which is an unholy xml mess.  Can you paste a link to the log where you saw that error, please?

svx:
There is an 'issue' with xml formating so it seems:

https://raw.githubusercontent.com/plone/training/master/export_code.rst starting from line 85

That could be also one possible reason why latex building on rtd is failing, but I am not sure about that

svx:
@fulv 
xml-issue

fulv:
There are actually two errors.  One is an indentation error of the code block around line 162.  

The other error is the one you saw.  It goes away if I remove the `:linenos:` thing.  Not sure why, because the second code block also has `:linenos:` and it works.  It must be because of that escaped inline xml.

I've just merged this fix:  https://github.com/plone/training/pull/148

PR #149, opened 2015-10-06
Fix some code mistakes in zpt samples, and typos.

fulv:


zpt.rst
```diff
@@ -68,9 +68,7 @@ Let's try it. Open the file ``training.pt`` and add:
 
 .. code-block:: html
 
-    <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en"
-          lang="en"
-          i18n:domain="ploneconf.site">
+    <html>
```
fulv: How about leaving all the attribute cruft out of the first example?  It's just so much cognitive load that distracts from the point of the code.

PR #177, opened 2015-11-23
Fix directions which led to duplicate resourcess being delivered

### LINKED ISSUE #174: Broken steps in theme-package.html#css-and-js-resources
This section tells to register resources twice: once in registry.xml and then again in manifest.cfg inside the implicit Diazo bundle:

* http://training.plone.org/5/theming/theme-package.html#css-and-js-resources

This is wrong and results in duplicate bundles output as seen in plone/Products.CMFPlone#1220. 

Trainers, please should choose one way or another and make the difference explicit.

davilima6:
(closes #174)

### DIFF
--- a/theming/theme-package.rst
+++ b/theming/theme-package.rst
@@ -34,7 +34,7 @@ Then we enable the virtualenv:
 Create a product to handle your Diazo theme
 ===========================================
 
-To create a Plone 5 theme skeleton, you will use mrbob's templates for Plone.
+To create a Plone 5 theme skeleton, you will use mr.bob's templates for Plone.
 
 
 Install mr.bob and bobtemplates.plone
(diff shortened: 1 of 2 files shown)


theming/theme-package.rst
```diff
@@ -462,15 +459,15 @@ As a starting point we use this rules set:
...
                </div>
                <div class="clearFix"></div>
              </div>
            </div>
-           <section class="row" id="viewlet-below-content-body">
+           <section id="viewlet-below-content-body" class="row">
              <div class="box">
                <div class="col-xs-12 col-sm-12">
-                <xsl:copy-of select="//div[@id='viewlet-below-content']"/>
+                <xsl:copy-of css:select="#viewlet-below-content" />
```
fulv: If one were picky, this should be `div#viewlet-below-content`.

theming/theme-package.rst
```diff
@@ -479,22 +476,22 @@ As a starting point we use this rules set:
      </replace>
 
      <!-- Left column -->
-     <rules if-content="//*[@id='portal-column-one']">
+     <rules css:if-content="#portal-column-one">
        <replace css:theme="#column1-container">
-           <div class="col-xs-6 col-sm-3 sidebar-offcanvas" id="sidebar">
+           <div class="col-xs-6 col-sm-3 sidebar-offcanvas">
```
fulv: Why did you drop the #sidebar?  (Other lines below, too.)

theming/theme-package.rst
```diff
@@ -536,14 +532,15 @@ Replace the placeholder with the real Plone top-navigation links:
      css:theme-children=".navbar-nav"
      css:content-children=".plone-navbar-nav" />
 
-Here we take the list of links from Plone and replace the placeholder links in the theme with it.
+Here we take the list of links from Plone and replace the placeholder links in
+the theme with it.
```
fulv: I know there are lots of places here where we don't follow it, but the rule for documentation is to use semantic linefeeds, not PEP8.  http://docs.plone.org/4/en/about/styleguide.html#id2

theming/theme-package.rst
```diff
@@ -726,15 +720,15 @@ bootstrap grid classes, we use an inline XSL snippet like this:
...
              </div>
              <div class="clearFix"></div>
            </div>
          </div>
-         <section class="row" id="viewlet-below-content-body">
+         <section id="viewlet-below-content-body" class="row">
            <div class="box">
              <div class="col-xs-12 col-sm-12">
-              <xsl:copy-of select="//div[@id='viewlet-below-content']"/>
+              <xsl:copy-of css:select="#viewlet-below-content" />
```
fulv: `div#viewlet-below-content`

theming/theme-package.rst
davilima6: The literal xpath > css translation would but the actual transformation will happen correctly in both cases and it's simpler to understand and demonstrate in a training, do you agree?

Under the hoods it seems that the only difference would be a tiny little bit of less performance but that should be really low and if it isn't we should work on that or put real disclaimers in the docs because as far as I'm concerned the CSS selectors should resemble real CSS selector, that is, best practices should apply (and avoiding specifying the tag when an id is unique is one of them).

theming/theme-package.rst
davilima6: As I mentioned [here](https://github.com/plone/training/issues/174#issuecomment-159054729), because of plone/plonetheme.barceloneta#53

fulv: Got it!

fulv: I see!

theming/theme-package.rst
davilima6: I actually agree but hadn't seen that so thanks! Most lines were already with 78/79/80 chars so I just followed that and fixed exceptions. I will do like this from now on, is that ok? (These docs are supposed to be completely rewritten still this year)

davilima6:
Thank you.

PR #181, opened 2016-02-26
Update vagrant instructions.rst to include BIOS virtualization note.

lukebrannon:
Had to enable virtualization in my BIOS before vagrant could be brought 'up.'  Added note to instructions.rst for specific case of ssh_exchange_identification resetting.


fulv:
Thanks!

PR #182, opened 2016-05-22
Move mastering plone to it's own folder

pbauer:


pbauer:
@svx Please review. The navigation still looks funny. Some parts of mastering plone have no left-navigation and I could not figure out why.

svx:
@pbauer you updated the version of sphinx, with the new version you have to configure that in the conf.py

```
# Theme options are theme-specific and customize the look and feel of a theme
# further.  For a list of options available for each theme, see the
# documentation.
html_theme_options = {
    "collapse_navigation" : False
}
```
This is how I did it, but I as I can see you tell sphinx in the conf.py to behave different depending if the docs are deployed to rtd or not, right ?

I did lots of more changes, locally now, because of the moving some settings and paths were messed up, creating some kind of loops and so on.
I will push these as a new branch and then you can have a look ?

One other point is that we should decide now what kind of content goes into sphinx and what kind/ which content goes on to the new static training.plone.org landing page, for example info about trainer, general overview [about all trainings] and so on.

For example I put trainer now on the static site. Which looks and feels at least for me way better, nicer and more friendly :)

pbauer:
The `on_rtd` is only needed for the theme to work outside and inside rtd. It does not change anything else. See https://github.com/snide/sphinx_rtd_theme#using-this-theme-locally-then-building-on-read-the-docs 
The list of trainers, upcoming trainings, overwrie and such should be in the static page, the trainigns themselves should be in sphinx so that new trainings can easily be added and outdated ones removed.
I'm anxious to see your branch. As long as it fixes stuff you can also simply push to this branch.

PR #185, opened 2016-06-24
Some tidying while reading

jean:


pbauer:
@jean Please do not put linebreaks into sentences. See http://docs.plone.org/about/documentation_styleguide.html#documentation-structure-styleguide 
The 80-character rule does not apply to docs.

pbauer:
@jean please rebase and remove the linebreaks.

jean:
@pbauer I like semantic linebreaks. See the discussion on https://github.com/apiaryio/api-blueprint/pull/324 for a PR where I was trying to introduce semantic linebreaks to a project (the relevant commits were squashed).

However semantic linebreaks is not about keeping a sentence on a single line. 
It is about breaking at clause or thought boundaries, and not reflowing unnecessarily. 
See http://docs.plone.org/about/rst-styleguide.html which is linked from "[follow this ReST styleguide and **use semantic linefeeds**](http://docs.plone.org/about/documentation_styleguide.html#documentation-structure-styleguide)". Note the linebreaks in the example there, and the following:

> manually insert line breaks after full stops, commas, or upon "grammatical" boundaries (and not merely word ones).

See also Brandon Rhodes's original post, itself linked from python.org: http://rhodesmill.org/brandon/2012/one-sentence-per-line/ where he responds to Limi, regarding the original title of the post:

> _Limi:_ Minor nit: When I first saw the headline of the article, I was confused — isn’t it more like “one idea” or “one concept” per line?
> > _Brandon Rhodes:_ You are quite correct, as usual, Alex — the title is a significant compromise, **given how many sentences themselves need to be split into several lines on a standard 80-column terminal.**

This goes back to Kernighan in 1974:

> Start each sentence on a new line. Make lines short, and break lines at natural places, such as after commas and semicolons, rather than randomly.

Semantic linebreaks are about using linebreaks for readability, ease of editing, for avoiding unnecessary reflowing (i.e. don't reflow a whole paragraph just to tidy things up), and for generating short diffs showing just what changed (diffs in the middle of long lines aren't great).

It's a matter of judgement, and I agree some of the breaks in my PR were ugly. I've pushed a commit that fixes a couple (and I'll try to improve more) .. but it also introduces some more linebreaks, which I think are improvements and that are in accordance with the intent of semantic linebreaks. 

You are of course free to prefer not breaking sentences across lines, but I ask that you then acknowledge that this diverges from the original formulation and the python.org style guide.
I hope you reconsider though and really go for semantic linebreaks :)

Anyway, now to properly read the rapido chapter ..

jean:
Thank you :bow:

PR #202, opened 2016-09-08
Fix a couple of bugs in the top5 blocks.

fulv:
A  couple of questions:
- where is Site Setup / Actions (I went into the ZMI to portal_actions)
- The search does not work for me.  If I put `context.app.records()` instead of `search` in the loop, the page displays my records, and the totals are `> 0`.  So I don't know why search does not return anything.

### DIFF
--- a/rapido.rst
+++ b/rapido.rst
@@ -814,7 +814,7 @@ We want to be able to sort the records according to their votes:
 
           def talks(context):
               search = context.app.search(
-                  "total>0", sort_index="total", reverse=True)[:5] 
+                  "total>0", sort_index="total", reverse=True)[:5]
               results = []
               for record in search:
                   content = context.api.content.get(path=record["id"])
--- a/rapido.rst
+++ b/rapido.rst
@@ -588,7 +588,7 @@ And if the element is a dictionary, we can access its items:
 
     <p>Average: {stats[avg]}</p>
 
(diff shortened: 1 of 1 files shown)


ebrehault:
- Site Setup / Actions is a new Plone 5.1 control panel (nicer than ZMI portal_actions, but does just the same). Maybe we should mention to go to portal_actions when using 5.0.
- regarding the search, I guess there is an index issue here, make sure `total` is declared as indexed in your .yaml and call @@rapido/the_app/refresh to reindex everything.

fulv:
I refreshed, and got this output:

```
Refreshed (id, total)
```

Still nothing in top 5.  Does it matter which .yaml file the index is declared in?  The training says to put it in `rate.yaml`, but since it's used by top5, it would make more sense to me to put it in `top5.yaml`.  Still, no matter which file I declare the index in, nothing happens.
I've added you as manager to my heroku test site, so you can take a quick look, if you don't mind.

ebrehault:
I found the problem.
In the `like` function, we need to call the record's `save` method:
```
def like(context):
    ...
    record.save(block_id='rate')
```
If we don't, the item is updated in the storage, but it is not indexed. That's something I haven't mentionned in the training (and neither in the Rapido tutorial).
Thank you for your review :)

ebrehault:
That's fixed.

PR #207, opened 2016-09-14
put all TTW things together

ebrehault:
@pbauer you asked me to put ttw theming + ttw dexterity + rapido + mosaic together, but then I realized there are also few other chapters which are about TTW, so I decided to move all of them in a single folder.

But maybe it is too radical, just tell me what you think.


pbauer:
That will not work since dexterity.rst is a integral party of the story of mastering plone. Since sphinx does not work with symlinks I don't know of a way to have the same file in two places but to include it in the toctree as `../mastering-plone/dexterity`

ebrehault:
Maybe the easiest solution is to create the TOC manually to insert the link to mastering-plone/dexterity.

pbauer:
exactly what I meant.

PR #208, opened 2016-09-17
Fix a couple of duplicate labels and an unmatched literal ending

jean:
Syntastic's Sphinx plugin picked up a couple of duplicate labels and
warned that ``for = `` should be ``for =`` for the closing marker to be
matched.

It doesn't look like either of the labels were being used.


jean:
Included in #209

PR #216, opened 2016-10-08
Apply Sphinx inline directives (file, kbd, Python domain, etc.)

jean:
This works toward making the docs more explicit and more amenable to
precise and consistent formatting.

I also fixed some typos, added some punctuation, capitalised some
acronyms, etc.

In some cases I left ``literal`` markup, where I couldn't decide what
kind of thing we were looking at. E.g. is a YAML "file" in a Rapido
app a file if it only lives in the ZODB? (I think it probably is.)

For changing Python "distribution" to "package", see
https://packaging.python.org/glossary/#term-distribution-package

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -53,6 +53,8 @@ This changelog is only very rough. For the full changelog please refer to https:
 
 - Exercises for behaviors_2, fix some emphasis [tschorr]
 
+- Fix some typos, apply inline directives: file, Python domain, GUI [jean]
+
 1.2.4 (2014-10-03)
 ------------------
 
--- a/deployment/ansible.rst
+++ b/deployment/ansible.rst
@@ -14,16 +14,16 @@ Installation
 Ansible is usually installed on the orchestrating computer -- typically your desktop or laptop.
 It is a large Python application (though a fraction the size of Plone!) that needs many specific Python packages from the Python Package Index (PyPI).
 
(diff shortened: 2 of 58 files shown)


jean:
I'm seeing some weird `duplicate label` warnings (the "other" file is the source file):

`mastering_plone/configuring_customizing.rst:4: WARNING: duplicate label customizing-label, other instance in /home/jean/repos/git/training/mastering_plone/configuring_customizing.rst`

There are also warnings about unknown options, I think that's OK ([ref](http://www.sphinx-doc.org/en/stable/markup/inline.html#role-option)).

`/home/john/repos/git/training/mastering_plone/buildout_1.rst:321: WARNING: unknown option: -v`

jean:
`/home/jean/repos/git/training/mastering_plone/restapi.rst:66: WARNING: Pygments lexer name u'http-request' is not known`

I think that's OK, it'll encourage us to look for such a lexer.

pbauer:
@jean please rebase. Would you mind using branches instead of forks for future changes? Building forks locally simply adds barriers and additional steps that I'm often to lazy to do.

pbauer:
Never mind fixed it locally and merged it.

jean:
@pbauer thanks!
I used a branch in my fork .. I didn't realise that a PR from a fork adds effort. 
I'll switch to using branches in the origin repo.

PR #222, opened 2016-10-15
fix typo in heading (boostrap)

stevepiercy:


stevepiercy:
I've signed the agreement at PloneConf, and sent it in via email.

p.s. -- Thanks for listening to my concern about the need to update the agreement documents and process.

pbauer:
Thanks!

PR #238, opened 2016-11-10
Move PloneConf 2016 to Previous Trainings

stevepiercy:


svx:
@stevepiercy we use rst not markdown :).

Also, the setup of sphinx is currently not configured for markdown and also the usage of md will make our test fail :)

stevepiercy:
@svx contributing.md is not part of the docs nor should be processed by Sphinx, but is used by GitHub as a special feature as described in this blog post: https://github.com/blog/1184-contributing-guidelines

It helps inform contributors who enter through GitHub how to contribute.

svx:
OK, but would you mind if I suggest that we do it, like we already doing it in other repositories ? See for example the setup we use on the docs:
https://github.com/plone/documentation/tree/5.0/.github

Reasons:
- unified setup for repositories
- less possible confusion for user 

But good point ! I will write a short part about that and will add it to the guideline part of the docs !

stevepiercy:
OK, wherever GitHub can find a contributing doc is fine, but the lack of a file extension makes it less human readable when served by GitHub. I suggest that it be renamed from:

https://github.com/plone/documentation/blob/5.0/.github/CONTRIBUTING

to:

https://github.com/plone/documentation/blob/5.0/.github/CONTRIBUTING.md

Would that be acceptable? I don't know how many repositories that would affect.

stevepiercy:
FTR, I found a more recent blog post where the `.github/` folder is mentioned.
https://github.com/blog/2111-issue-and-pull-request-templates

I think that if the file contains markdown syntax, then it should have the `.md` file extension. I'll update this PR accordingly.

PR #245, opened 2016-11-14
Fix miscellaneous Sphinx warnings on clean build

stevepiercy:
These are only a few of the warnings I was able to fix. There are dozens more, but I couldn't figure out how to get them to pass.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,9 @@ This changelog is only very rough. For the full changelog please refer to https:
 
 1.2.5 (unreleased)
 ------------------
+- Update README.rst to refer to how to build the docs locally.
+  [stevepiercy]
+
 - Add CONTRIBUTING.md
   [stevepiercy]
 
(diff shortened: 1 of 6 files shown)


pbauer:
I think the travis-build will always fail with the openssl-error when trying to build a pull-requests from forks. Try using a branch next time.

svx:
yes !

stevepiercy:
@pbauer to clarify, should I create a branch off of `plone/master`, e.g., `plone/my-branch`, then push commits onto, and make PRs from, that branch?

In Pylons Project, we don't allow non-core contributors to make branches on `Pylons/project_name`, but it seems that Plone allows more liberty to do so.

I am still on-boarding and trying to learn the ropes in Plone.

I have a couple of upcoming PRs, so I will adapt accordingly.

pbauer:
Yes, you can go head an create branches. The branches will be deleted after merging anyway and it is much easier for everyone to quickly fix something in a branch, whereas it is usually quite a hassle to do that in a fork. I'd have to fork your fork and and make a pr from my fork to your fork...

branches rule!

PR #250, opened 2016-12-13
Update relations.rst

zmijunkie:
just a minor typo: together is used twice

### DIFF
--- a/mastering_plone/relations.rst
+++ b/mastering_plone/relations.rst
@@ -1,7 +1,7 @@
 Relations
 =========
 
-You can model relationships between content items by placing them in a hierarchy (a folder *speakers* containing the (folderish) speakers and within each speaker the talks) or by linking them to each other in Richtext-Fields. But where would you store a talk that two speakers give together together?
+You can model relationships between content items by placing them in a hierarchy (a folder *speakers* containing the (folderish) speakers and within each speaker the talks) or by linking them to each other in Richtext-Fields. But where would you store a talk that two speakers give together?
 
 Relations allow developers to model relationships between objects without a links or a hierarchy. The behavior :py:class:`plone.app.relationfield.behavior.IRelatedItems` provides the field :guilabel:`Related Items` in the tab :guilabel:`Categorization`. That field simply says ``a`` is somehow related to ``b``.
 


gforcada:
@zmijunkie thanks for the pull request! Would you mind writing a minimal entry on ``CHANGES.txt``? Forget about the Contributors agreement (although it does not hurt to sign it if you want).

pbauer:
We don't need a changenote for typos and other minot changes.

pbauer:
The build only fails because the ssl-thing cannot deal with forks. I'll merge.

svx:
I am really sorry that the *ssl-thing cannot deal with forks* ....

This is on purpose since we tell travis to push after a successful build to training.plone.org with is done via a ssl hook/with a looked down key for a special user for one special task.
This key is owned by this repo and if you fork, the permissions will change, so it is completely OK that the build is failing because of wrong ssl permissions.
If it don't, it would be really scary, because that means that everyone who forks and makes sure that some basic sphinx tests are running through would be able to publish whatever they like on docs.plone.org.
Yes, I agree it is still not perfect and personally I would like to change the current setup completely, because it has its challenges and at least for me, it is still not secure enough.

The thing is I want to avoid things over and over again, so maybe it is time to thing about how to do training.plone.org in a proper and nice way, but then again hosting it, is only one small part of it.
IMHO there is way more work to do and it would be nice to have first a *master plan* otherwise we will keep doing things over and over again. 

If you have a better solution for the current setup I am looking forward to see that ! 

I apologize if that was now maybe too harsh, this is not meant to be personally, but there are sometimes reasons why stuff is setup in a certain way even if it maybe not perfect.

pbauer:
I see no real problem with the setup, I simply try telling people to use branches instead of forks. 

But: Is it be possible to change the order of the build so that the rst-build and linkcheck-test  is run before the "ssl-thing" fails? This way I will agt least have a indication if the build of the master will fail or not after merging. Currently it fails since www.zope.org is down an the link to http://www.zope.org/the-world-of-zope. Also mr.roboto always complains about missing changelogs even though that is not needed fot the training-docs. Can we disable that?

Issue #240, opened 2016-11-10
Fix Travis-CI builds


stevepiercy:
I have no clue how to fix this:
https://travis-ci.org/plone/training/builds/174755889#L243

```bash
The command "openssl aes-256-cbc -K $encrypted_129e1d4a0f4f_key -iv $encrypted_129e1d4a0f4f_iv -in deploy_key.enc -out $HOME/.ssh/id_rsa -d" failed and exited with 1 during .
```

svx:
@stevepiercy you can't .
This is an issue with Travis itself, this happens every time if there is/was an openssl/libressl update.

Usually it takes a couple of days, till Travis updates to their container images.
This is really annoying and this is one of the reasons why I want to switch CI away from Travis.

stevepiercy:
@svx if this is being discussed elsewhere, then feel free to close this issue.

If not, then what are some options?

svx:
The easiest one would be to change from openssl to gpg.

Thing is, I am busy working on setup2.0 for training and docs and with setup2 we will use a completely different approach of testing and deploying, so I personally prefer to focus currently on setup2.0 if someone else would like to fix this issue, awesome please do ! :)

stevepiercy:
@svx that sounds like a good plan. Where is setup2.0? I can be a test subject or contribute.

stevepiercy:
We use GitHub Actions now. Closing.

PR #255, opened 2017-03-29
changed loot to lot

giammi:
I, Gian-Maria Daffré, agree to have this published under Creative Commons 4.0 International BY


pbauer:
@giammi the build only failed since the trainings docs cannot build correctly with forks (braches are fine though). See #240

PR #265, opened 2017-06-27
Remove notes on features.rst that are no longer valid.

robinsjm2:
Dexterity is now supported for Working Copy Support and Placeful Workflow.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,6 +6,7 @@ This changelog is only very rough. For the full changelog please refer to https:
 1.2.5 (unreleased)
 ------------------
 - Update Features training to reflect that Dexterity is now supported by Working Copy Support and Placeful Workflow.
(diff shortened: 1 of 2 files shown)


tkimnguyen:
Does this need to be merged into 5.0 and 4.x branches?

hvelarde:
yes, indeed…

tkimnguyen:
@robinsjm2 according to https://docs.plone.org/about/contributing.html#documentation-for-different-versions-of-plone "Documentation changes that are valid for multiple versions of Plone can be done by making multiple pull requests, or by cherry-picking which may be easier to do when branches are widely different. When you are not sure, note in your pull request that you think this is valid for other versions of Plone as well, and the documentation team will take care." and see https://docs.plone.org/about/cherrypicking.html

PR #267, opened 2017-06-28
Add note on known limitations concerning Dexterity / Working Copy Support.

robinsjm2:
After removing the note that mentioned Working Copy Support does not work with content types created via Dexterity, there was some discussion that brought to light the known limitations by name.  This updated note does claim support but also refers directly to the known limitations, providing better information for the reader to choose how best to use the feature.


hvelarde:
your fork is outdated; please rebase and include a link to the open issue:

https://github.com/plone/Products.CMFPlone/issues/665

PR #266, opened 2017-06-27
Fix redirects

hvelarde:

PR #268, opened 2017-06-28
Use correct links for ZCA itself and usage in Pyramid

stevepiercy:
In #266 links to ZCA were updated, but since then I realized two things.

1. The canonical URL for docs is https://zopecomponent.readthedocs.io/en/latest/ per https://github.com/zopefoundation/zope.component/pull/25
2. Under the section about Pyramid, the context was for how Pyramid uses ZCA, not for ZCA itself, and should in fact point to Pyramid docs at https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/zca.html

This PR corrects the URLs and makes them consistent across the translation files as well.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,10 @@ This changelog is only very rough. For the full changelog please refer to https:
 
 1.2.5 (unreleased)
 ------------------
+- Use correct links for ZCA itself and its usage in Pyramid, as well in
+  translation files.
+  [stevepiercy]
+
 - Update Features training to reflect that Dexterity is now supported by Working Copy Support and Placeful Workflow.
 
 - Fix a typo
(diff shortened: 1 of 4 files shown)


stevepiercy:
I'm not sure why the build fails on Travis. Anyone got an idea?

pbauer:
travis always fails with the training-docs when using a fork instead of a branch. @svx will deal with that at some point. until then simply ignore it.
The error is always:

```
The command "openssl aes-256-cbc -K $encrypted_129e1d4a0f4f_key -iv $encrypted_129e1d4a0f4f_iv -in deploy_key.enc -out $HOME/.ssh/id_rsa -d" failed and exited with 1 during .
```

stevepiercy:
OK, I probably should make a branch off master even if it is a minor docs fix. Sorry about that.

pbauer:
No problem, you can commit small fixes directly to master. Pull-requests should best be from branches. We will delete them on merging anyway.

PR #188, opened 2017-09-20
Update Vagrant Build to use Ubuntu 16.04 LTS

tschorr:
The current Vagrantfile uses Ubuntu 14.04 LTS, but current Ubuntu LTS is 16.04 (Xenial Xerus).


tschorr:
@svx thanks for the reminder :-)

tschorr:
Some notes:

- there is a corresponding branch for the training_buildout also named `vagrant_xenial_update`
- for now I have commented out the download + unpacking of the buildout_cache, since the buildout is now using 5.0.7 and there is now buildout cache tarball for this release
- Travis build failure seems unrelated (openssl complaining about undefined iv)

tschorr:
... and in case anybody wonders about the messed up branch names, I accomplished this by doing `git-hub pull attach` to turn the issue into a PR.

plone_training_config/Vagrantfile
```diff
@@ -22,6 +22,16 @@
   # A workaround for missing symbolic links on windows:
   # config.vm.synced_folder "eggs/", "/home/vagrant/buildout-cache/eggs/"
 
+  # Read initial package index information + upgrade
+  config.vm.provision :shell, inline: <<-SHELL
+      apt-get update
+      apt-get -y upgrade
```
pbauer: isn't that done by aptgetupdate.pp?

plone_training_config/Vagrantfile
```diff
@@ -22,6 +22,16 @@
...
+  SHELL
+
+  # We need to install puppet on the guest before we can use it
+  config.vm.provision :shell, inline: <<-SHELL
+      apt-get install -y puppet
```
pbauer: why did we not need this before?

plone_training_config/manifests/plone.pp
```diff
@@ -1,28 +1,28 @@
 class plone {
 
-    $plone_version = "5.0.6"
+    $plone_version = "5.0.7"
```
pbauer: I'd prefer to use 5.1rc1 for the training. Also 5.0.8 is already out and 5.0.9 is pending.

plone_training_config/Vagrantfile
tschorr: chicken or egg: `aptgetupdate.pp` needs the puppet provisioner, so puppet has to be installed on the guest. It isn't in the `ubuntu/xenial64` base box.

plone_training_config/Vagrantfile
tschorr: same as above

plone_training_config/manifests/plone.pp
tschorr: The `plone_version` in the manifest is only used for the buildout cache download. Imo it should match the version in https://github.com/collective/training_buildout. So what we need to change is the training buildout which currently uses 5.0.7.

PR #279, opened 2017-10-02
Fix some warnings and errors

jean:
Uniquify IDs that are computed from toctree caption.
Inconsistent indentation caused some spaces to be included in filenames.


jean:
Superseded by #282

PR #282, opened 2017-10-05
Git rid of "duplicate label" warnings; add some markup

jean:
Uniquify IDs that are computed from toctree caption.
Inconsistent indentation caused some spaces to be included in filenames.


jean:
Thanks! What's up with that openssl command failure?

pbauer:
The build never works with forks. Only with branches.

PR #289, opened 2017-10-08
Fix some code-block types, indentation, label warnings

jean:


ttw/ttw-advanced-2.rst
```diff
@@ -1 +1 @@
-.. include:: ../theming/ttw-advanced-2.rst
```
svx: I am not sure this is a good idea :)

Using ``:include`` is tricky, the tone, voice and audience may differ also the style of writing.
This can lead to weird and confusing documentation.

Even more important this will break in the future if people try to build only the ttw docs.
It may work now, but not in the future.

jean: @svx I tend to agree with you. I got this change via a merge. What's the solution, though? Duplicate the chapter and recommend diffing to keep it in sync? We have the same issue with two `dexterity.rst` files which seems to be pretty much identical except for formatting.

svx: I usually duplicate and change the file then to my needs, like I make sure to use the same style-guide, voice of tone and same rst formating and all such stuff.
Including a file with a different rst formating, like headers for example can produce some 'fancy weird docs'  :)

Also I check if the copied file is OK in terms of content, do I have everything ? Does the user knows all the terms, I had it some time in the training docs, that ppl used ``include`` and suddenly they had terms which were never used before in this training in the docs, because of the include.

One other point is we do not know if content will change/move or will be renamed. 
This can lead to broken docs or missing content :)

pbauer:
@jean please use a branch instead of a fork because the travis-build will always fail otherwise.

jean:
@pbauer ahhh that's why, sorry, sounds like something I should know about ...

pbauer:
please rebase and I will merge. and next time use a branch 😄

jean:
Superseded by https://github.com/plone/training/pull/293

PR #297, opened 2017-10-11
Angular fixes

ebrehault:


ebrehault:
@fulv by the way, I tested the exercise with the last plone.restapi-angulat (1.2.4) and it works fine

PR #317, opened 2017-10-17
Fix authentification angular documentation

bsuttor:
Training uses this.services instead of this.plone variable.


ebrehault:
Sorry I just fixed it :)

PR #326, opened 2018-04-09
Screenshots

brendaavdijaj:


mastering-plone/add-ons.rst
```diff
@@ -17,6 +17,11 @@ Some notable add-ons
 `collective.easyform <https://pypi.python.org/pypi/collective.easyform>`_
   A form generator and the successor to `Products.PloneFormGen <https://docs.plone.org/develop/plone/forms/ploneformgen.html>`_
 
+.. image:: _static/add_ons_easyform_1.png
```
pbauer: Indent this section same as the text to make it more obvious that it belongs to the easyform addon. 

If you add a `:scale: 50%` the image will be additionally rendered as a link to the full size. Please always use a scale!

You should always use `.. figure::` instead of `image`. `figure` allows you to add a caption and legend to the image. See http://docutils.sourceforge.net/docs/ref/rst/directives.html#figure. In our case we at least sometimes want to add a caption.

mastering-plone/add-ons.rst
```diff
@@ -17,6 +17,11 @@ Some notable add-ons
...
+
+.. image:: _static/add_ons_easyform_2.png
```
pbauer: Caption: "Editing a form field through the web."

mastering-plone/add-ons.rst
pbauer: The caption here could be: "A simple form created with collective.easyform"

mastering-plone/configuring_customizing.rst
```diff
@@ -82,6 +82,8 @@ Let's change the logo.
 * Go to http://localhost:8080/Plone/@@site-controlpanel
 * Upload the Logo.
 
+.. image:: _static/configuring_customizing_logo.png
```
pbauer: It would be great to use a better logo that the old one from 2014. One that we can keep for a couple of years 😄

mastering-plone/features.rst
```diff
@@ -42,13 +42,19 @@ Some commands you will use rather often are::
 
     A standard installation listens on port 8080, so lets have a look at our Zope site by visiting http://localhost:8080
 
+	.. image:: _static/features_plone_running.png
+		:scale: 50 %
```
pbauer: The indention of the options is a little much. It is still valid rst I think but I think 4 spaces should be enough.

svx:
Please also use
``
:alt: alternate text
``

So that people who use screen-readers know what your image is about

mastering-plone/configuring_customizing.rst
brendaavdijaj: Is it possible to change the size of the logo? I changed it but I think it's to big...

PR #351, opened 2018-11-02
Use a period at the end of the sentence.

stevepiercy:


tschorr:
@stevepiercy sorry but I don't understand.

Section 2.1 in the linked document is about "Introducing the text that follows" not lists. And I can't see a conflict with the Plone docs styleguide: Every sentence is already on a single line and the example in the styleguide even contains a colon (followed by a list in this case). 

What is it I am missing?

stevepiercy:
A sentence must end with a period, not a colon, in this context. Lines 75-80 form a single paragraph, telling the history of evolution of CGI to Zope and leading onward. It is not a list.

PR #371, opened 2018-11-08
better setup --> requirements + constraints

loechel:
change setup so that github is not complaining about security issues and it is actual more the way we do in Plone


loechel:
ok tests are failing due to linkcheck problems, which is unrelated to my pull rerquest.

constraints.txt
```diff
@@ -0,0 +1 @@
+Sphinx >= 1.7.1, <= 2.0.0
```
stevepiercy: Is this what Plone uses? I don't understand (1) why lower the minimum version number to less than what it was in `requirements.txt`, and (2) allowing the 2.0.0 version to be used.

1. Sphinx is currently at 1.8.1. Suggest either that as the minimum or 1.7.6 as previously used.
2. Use `< 2.0.0` to avoid a major upgrade with potentially breaking changes.

loechel: You are totaly right, seems that I am to tiered and making stupid mistakes.

PR #375, opened 2018-12-19
Remove a reference to an old file name, used by AT.

ramiroluz:


ramiroluz:
@pbauer added you as a reviewer because I am not sure if this is a correct or appropriate PR.

stevepiercy:
Also when I looked at where the source files should be, this template was missing:
https://github.com/collective/ploneconf.site/tree/zpt_2/src/ploneconf/site/browser/templates

PR #406, opened 2019-03-08
WIP: Plone 5.2 and Python 3

pbauer:


stevepiercy:
@pbauer I started a review, but realized that it would be easier and faster to just make changes and push commits. I could either push commits to `plone52` so they get into this PR, or create a new branch, push commits to that, and submit a PR to `plone52`. Do you have a preference? Please let me know.

pbauer:
This is dependent on https://github.com/collective/training_buildout/pull/17 and not ready for review. 

First we'll have to discuss what we want to take from https://github.com/collective/ploneconf.site/pull/3 as well because that would result in a lot more changes.

PR #414, opened 2019-05-23
Fix react and volto bootstrapping

jensens:
instruction were not working, 
- fix nvm download url and use latest
- use node lts
- fix yarn install to be universal and make it work with nvm
- use `npx`


tisto:
LGTM

PR #411, opened 2019-04-04
Fix #410

stevepiercy:
This PR is an almost complete fix for #140, except for https://github.com/plone/training/issues/410#issuecomment-479814160

I'll push commits to this PR once I have an answer to that comment. Otherwise I made a guess as to what should be done with the files.

PR #415, opened 2019-05-24
Update to py3 and fix errors

jensens:


stevepiercy:
#411 has some similar changes and would introduce conflicts. If you merge my #411 into your branch, then push, then we can close #411 in favor of this PR.

jensens:
ah, sh** got no update, will merge all missing, stay tuned.

PR #419, opened 2019-08-17
fixes recommended by @stevepiercy

cdw9:


stevepiercy:
@cdw9 can you check this build error? Looks like a missing file that I did not catch.
https://travis-ci.org/plone/training/jobs/573392831#L1622

cdw9:
I pushed 2a295eb to fix that link. Thanks for your review @stevepiercy !

PR #428, opened 2019-09-06
Improve deploy.sh

ericof:
Update deploy script to handle errors during rsync and properly fail the build


ericof:
@svx If, for some reason, rsync does not work the script will exit with status code 1. This way we will not miss warnings something is wrong on the deployment server.
Important: This improvement is not to fix an issue with the rsync, but just to raise an error if an error happens.

PR #450, opened 2019-10-21
Add info to create your own package

Gomez:
As noted in the training


stevepiercy:
This is already covered in the [training doc](https://github.com/plone/training/blame/master/voltohandson/quickstart.rst#L42), but I think this could be made more visible for readers who rapidly scan through things.

PR #458, opened 2019-10-21
Testing training: geckodriver clarification

tkimnguyen:


testing/start.rst
```diff
@@ -58,3 +58,6 @@ Let's run all tests, including robot tests (we will cover these later):
 .. code-block:: console
 
   $ plonecli test --all
+
+.. note::
+  If you get an error about missing ``geckodriver``, you will have to install ``geckodriver``. On macOS, you can use ``brew install geckodriver`` then re-run ``plonecli test --all``.
```
jensens:
```suggestion
  If you get an error about missing ``geckodriver``, you will have to install ``geckodriver``. 
  On macOS, you can use ``brew install geckodriver``. 
  On Ubuntu, you can use ``apt install firefox-geckodriver``.
  Then re-run ``plonecli test --all``. 
```

jensens: one sentence per line.

tkimnguyen: thx!

PR #461, opened 2019-10-26
Remove mention of removed logLevel option

datakurre:
I just released gatsby-source-plone 0.2.0 where logLevel option has been removed, because we replaced custom logger with gatsby reporter API https://www.gatsbyjs.org/docs/node-api-helpers/#reporter


datakurre:
Completely unrelated issue is that we probably need to rewrite this tutorial with simpler use case before next trainings. Trying to reproduce complete Plone site has been good experiment on proving that our plugin works, but may not be the simplest or most common use case.

Issue #475, opened 2020-02-14
Update solr training


reinhardt:
The solr training is out of date and doesn't work as described any more, e.g. because collective.solr uses kitconcept.recipe.solr instead of collective.recipe.solrinstance now.

reinhardt:
I'll try looking into it today.

reinhardt:
First step was done in #476. The first couple of chapters have mostly been brought up-to-date, up to and including "Solr Buildout Configuration". The remaining chapters still have some references to c.recipe.solrinstance. In the setup part, `bootstrap-buildout.py` should be replaced with `venv` and `pip install`.

stevepiercy:
@reinhardt we've migrated from using rST to MyST. We would accept a PR that addresses this issue.

FWIW I recently updated all the links per the split between Lucene and Solr.

stevepiercy:
Closing as won't fix. The solr training is now archived.

PR #476, opened 2020-02-14
Update Solr training

reinhardt:
The part in the setup where I download half a dozen files with curl is not particularly pretty, but I can't think of a much nicer way short of cloning the whole repository (which seems like overkill). Suggestions welcome, but it at least it works now.

There are some outdated bits left. I hope to find some time to update them soonish.

Refs #475

### DIFF
--- a/solr/setup.rst
+++ b/solr/setup.rst
@@ -44,8 +44,8 @@ Create a buildout (*buildout.cfg*) which installs both requirements
     solr-config = config
 
     [versions]
-    collective.solr = 6.0a1
-    collective.recipe.solrinstance = 6.0.0b3
+    collective.solr = 8.0.0a5
+    kitconcept.recipe.solr = 1.0.0a5
 
 
 Run buildout:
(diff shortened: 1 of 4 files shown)


reinhardt:
@balavec, if you want you can try the "Setup" part already. I'm still polishing some of the rest.

reinhardt:
There are still some outdated bits, but I don't know when I'll get around to continue working on this. Let's merge it, at least the setup part should work now.

solr/setup.rst
```diff
@@ -27,23 +31,28 @@ Create a buildout (*buildout.cfg*) which installs both requirements
...
+    collective.solr = 8.0.0a5
+    kitconcept.recipe.solr = 1.0.0a5
 
 
 Run buildout:
 
 .. code-block:: console
 
-   python2.7 bootstrap-buildout.py
+   python bootstrap-buildout.py
```
tomgross: Is this still state of the art, or virtualenv / pip?

reinhardt: Ah, you're right, that should probably be updated as well in the next round. Thanks!

PR #483, opened 2020-10-31
better explanation of variables and references

pbauer:


jensens:
If at it, better use word "symbolic links" and - from my experience - do not expect everyone to know about them. So I would link to the wikipedia article https://en.m.wikipedia.org/wiki/Symbolic_link

PR #486, opened 2020-11-30
Removing bind and adding a functional component in react

iFlameing:


pbauer:
@iFlameing I don't have time to review this and also not enough deep react knowhow. 
So I trust you that the change is ok. Please revert if I'm wrong.

PR #502, opened 2021-05-14
Update dexterity_reference.rst: wording datagrid field

ksuess:


jensens:
LGTM. While at it you could convert to one sentence per line. But just as an idea, not as an todo.

PR #522, opened 2021-09-10
Fixes after theme: toc in main navigation, remove custom.less

ksuess:


ksuess:
This brings back warnings. But I think they are valid warnings. The chapters will be removed anyway.

`/docs/mastering-plone/zpt.md: WARNING: document isn't included in any toctree`

docs/index.md
```diff
@@ -127,28 +127,3 @@ glossary
 {doc}`teachers-training/index`
 
 : How To Give Technical Trainings
```
stevepiercy, line 129: If we remove these toctree entries, and we will remove the files eventually, why not remove the files now?

ksuess: Some are referenced, like code (info about checking out branches of the training code).
Some should be removed. This could be done while working on content next days. > @pbauer 
teachers training: While adding authors instructions to markdown, I'll add a reference in second toc in main navigation.

PR #513, opened 2021-09-01
WIP: move info about trainings in general to /about

ksuess:
- markdown /MyST instructions for authors
- instructions to setup sphinx, etc locally
- how to prepare and give a training: communication

PR #524, opened 2021-09-10
Add info material for authors and trainers and add navigation entries

ksuess:


ksuess:
It's a first step.

Further valuable info to be included:
- some more info about code snippets: https://github.com/plone/training/pull/513#issuecomment-912169849
- info about linkcheck
- authors please check warnings before merging

PR #526, opened 2021-09-11
Contributing review linkcheck

stevepiercy:


ksuess:
Would you rebase, please, thanks.

stevepiercy:
OK, rebased, and now this PR is ready for review.

stevepiercy:
I noticed that there is a some copy-paste from the two Mastering Plone trainings into the Contributing docs. If we are going to duplicate content, we should use `include`. Copy paste is not maintainable.

I also don't want to duplicate existing external documentation. I think we went a little too far with the MyST examples, and we should just pull back and delete them, except when they add clarity or something that does not exist in the official MyST docs.

ksuess:
Thanks for extending the info! 
I also think now that the /contributing/markup-syntax is not really necessary. It took me some time to find the several markddown rules and directives in markdown, myst and sphinx docu, but it's there. So, how you say, the external docu should not be duplicated here. +1 for deleting. The only bit that could be included somewhere else is the request to add alternative text to images for accessibility reason.

ksuess:
And yes "Building the documentation" should now be in /contributing and nowhere else. I removed it from mastering 6 but not from mastering 5. This needs to be done.

stevepiercy:
> The only bit that could be included somewhere else is the request to add alternative text to images for accessibility reason.

Yes. As well as any other style that we want to either enforce as required, or suggest as a best practice. See #527.

stevepiercy:
> And yes "Building the documentation" should now be in /contributing and nowhere else. I removed it from mastering 6 but not from mastering 5. This needs to be done.

See #528

ksuess:
**teaching**:

> Trainers should read {doc}`setup-build` and the trainings in {doc}`/plone_training_config/instructions` and {doc}`/teachers-training/index`.
> These documents help trainers prepare for a successful training experience.
> 

could go from /contributing/trainers to /teachers-training/index
and than delete /contributing/trainers from /contributing as this is meant for contributing not teaching. 
Everything concerning teaching can stay in /teaching: 
- an intro
- the existing teachers training.

I think /teachers-training/index.html#complete-mode-and-compact-presentation-mode can be replaced with your way better explanation of modes /contributing/setup-build.html#available-documentation-builds

my /teachers-training/index.html#setup-environment is superfluous with the added paragraph mentioned above (Trainers should read {doc}`setup-build` …

docs/contributing/authors.md
```diff
@@ -0,0 +1,270 @@
...
+
+
+## Synchronize the Browser While Editing
+
+Use gulp to view changes in the browser while editing documentation.
+
+Install the gulp command line utility.
+
+```shell
+npm install --save-dev gulp-cli
```
ksuess, line 255: This is not needed with adding gulp-cli to package.json

docs/contributing/authors.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+Install the gulp project.
+
+```shell
+npm install --save-dev
```
ksuess, line 261: just:
`npm install`

ksuess (changes requested):

ksuess:
@stevepiercy as you investigated a lot in linkcheck and check of spelling: Do you see a way to check if {figure} and {image} have alternative texts / alt attributes?

docs/contributing/authors.md
stevepiercy: Do we care whether we install dependencies into a production environment versus development? If so, then we should use `--save-dev`.

I guess it does not matter because we do not deploy any node modules to the production server anyway. I'll fix this.

docs/contributing/authors.md
stevepiercy: Ah, yeah, but it got added because I did that. :wink:

I'll remove this one.

stevepiercy: Fixed in c39f9f5d363b50bc007264b6426ecec63b5bb770.

docs/contributing/authors.md
stevepiercy: Fixed in c39f9f5d363b50bc007264b6426ecec63b5bb770.

stevepiercy:
Somehow my version of `markup-syntax.md` got obliterated when I did a rebase a couple days ago. I was able to restore it. I'd like to keep the examples because they suggest best practices that are specific to Plone Training documentation (and they are hard to find).

I also realized while writing it that it needed to become a more generalized docs writing guide. I renamed and refactored it accordingly.

I'm working on the next comment about moving teacher stuff around. That's a great idea!

stevepiercy:
@ksuess I've completed revisions per your feedback, including https://github.com/plone/training/pull/526#issuecomment-918838702. This is ready for another review.

ksuess:
Please correct me if I'm wrong, but I think now that we have /contributing and / teaching the one page /contributing/trainers can be deleted.

ksuess:
I would say with /contributing/trainers removed, it's good to merge.

ksuess (approved):

spereverde (approved):
wow, thanks for all the work and updates!
I checked it out locally and followed the new docs to build everything - all worked fine :-)

pbauer (approved):

stevepiercy:
> Please correct me if I'm wrong, but I think now that we have /contributing and / teaching the one page /contributing/trainers can be deleted.

Quickdraw @pbauer merged the PR before I could remove it. I'll create a new PR for that.

stevepiercy:
Actually, it appears that deleting `contributing/trainers.md` would [delete content that does not exist elsewhere and should be moved into an appropriate location](https://github.com/plone/training/blame/866fbfe0425ab53755bd203695e5d58a85aafa50/docs/contributing/trainers.md#L18-L49). I'll create a new issue.

PR #538, opened 2021-09-22
Add OpenGraph meta data

ksuess:


stevepiercy (approved):

stevepiercy:
virtualbox.org has been a problem for me, too. It was down for several hours when I first added linkcheck. If it goes down again in the future, I think it is OK to ignore it. It's not going away, like a blog post from 2016.

This looks good to me!

ksuess:
This PR does not add (meta data) content with the intent not to mix technicals with content.

stevepiercy (approved):
Looks goood to me.

The calling of a Python file from `Makefile` is a new technique to me.

PR #541, opened 2021-09-24
Change from {doc}`glossary` to {ref}`glossary-label` because it was l…

stevepiercy:
…inking to the official Python 3 docs.


stevepiercy:
I'm going to merge this as soon as it passes the build. It is too minor for review, but it is important to note that links from intersphinx configuration might override local links.

PR #564, opened 2021-11-10
Move Makefile to root: Run make from root

ksuess:
Enhanced help message: make help


Makefile
```diff
@@ -1,11 +1,195 @@
...
+#
 
-help:
-	@echo "Please change into the directory docs/ before running any make commands"
+# You can set these variables from the command line.
+SPHINXOPTS    = 
+SPHINXBUILD   = sphinx-build
+PAPER         =
+DOCS_DIR      = ./docs/
+BUILDDIR      = ../_build/
```
stevepiercy, line 9: Can you verify the `BUILDDIR` is correct, now that the Makefile has moved up a level? I think when the docs are deployed, its location matters.

stevepiercy: @ericof might know, or it might be in the GitHub Actions workflow.

ksuess: Stupid me moved and modified in one commit. So it is not obviously that I changed the BUILDDIR. The build directory _build is, same as before, in /.

stevepiercy: Here it is:

https://github.com/plone/training/blob/695659fb79d92373664fe805d2f08e9ffbf7bde5/.github/workflows/build_deploy.yml#L44

stevepiercy: If the builddir is still in the same location as before, then we're good.

ksuess: No, i did not change. But as the dir is changed, than build is run, the _build directory is in place.

stevepiercy: Got it. Please merge.

stevepiercy (approved):

PR #568, opened 2021-12-05
Fix missing_value for 'Set' field.

ksuess:


stevepiercy:
@ksuess does this need review? I haven't run through this training, so I'm not deeply familiar with it and don't feel qualified. Maybe @pbauer? If you're happy with it, then please merge.

PR #596, opened 2022-03-15
Feautre: Multiple spelling issues fix [2]

avimishra18:
See: #594

PR #594, opened 2022-03-13
Feautre: Multiple spelling issues fix

avimishra18:
# Features
This Pull Request solves the following issues relating to the `make spellcheck` command. 

1. Reduced ~25% of *misspelled words* from 1296 to 956 by:
   1. Adding words to `spelling_wordlist.txt`
   1. Corrected spellings in some file
1. Possibly Fixed All: `WARNING: image file not readable: <filename.ext>`. Including but not limited to these files: 
   1. [Customize CSS/Less Variables](https://training.plone.org/5/theming_plone_5/basic.html#customize-css-less-variables)
   1. [Exercise 3 - The <drop> Directive](https://training.plone.org/5/theming_plone_5/ttw-advanced-2.html#exercise-3-the-drop-directive)
   1. [Exercise 1 - Create A New Theme That Inherits From Barceloneta](https://training.plone.org/5/ttw/ttw-advanced-2.html#exercise-1-create-a-new-theme-that-inherits-from-barceloneta)
1. Removed all *misspelled words* from the following training: 
   1. React
   1. Volto

[Community Post Link](https://community.plone.org/t/information-for-interested-gsoc-students-2022/14893/31?u=avimishra18)

PR #600, opened 2022-03-15
Feautre: Multiple spelling issues fix [3]

avimishra18:
See:
- #594
- #596

### DIFF
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -16,7 +16,6 @@ backlinked
 backlinks
 backticks
 balancer
-bar
 Barceloneta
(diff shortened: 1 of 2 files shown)


stevepiercy:
Please incorporate all the comments I made about updating docs in the original PR: https://github.com/plone/training/pull/594 Assume that none of the comments are resolved.

Whenever you remove an item from `spellcheck_wordlist.txt`, you must run `make spellcheck` to see whether it shows up as a spelling error in the docs. If it does, then you need to correct the docs by either making acronyms upper-case, such as "cli" -> "CLI", or if it is an inline literal, then surrounding it with backticks.

PR #614, opened 2022-04-20
No longer require virtual environment activation

stevepiercy:
This PR makes it no longer required to activate the virtual environment by changing `make build` to `make bin/python`. It also updates the Contributing documentation. Finally this is now aligned with the main Plone 6 Documentation Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,8 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    = 
-SPHINXBUILD   = sphinx-build
+SPHINXBUILD     = $(realpath bin/sphinx-build)
+SPHINXAUTOBUILD = $(realpath bin/sphinx-autobuild)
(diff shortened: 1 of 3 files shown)


stevepiercy:
The failling linkchecks are for the Plone Demo, which can be ignored for now.

ksuess (approved):
Just cosmetics. The console output is nearly correct, the _build directory is ./_build not ../_build

stevepiercy:
> Just cosmetics. The console output is nearly correct, the _build directory is ./_build not ../_build

That would be a change outside this PR scope. That would put the build dir inside `/docs/` instead of the project root. That would also deviate from `plone/documentation`, and start placing builds inside what is documentation source code. If we change it here, we should change it there. I am would be -1 on the change, especially because I like to select the `docs` dir when doing a regular expression search across files.

PR #628, opened 2022-07-24
Fix tags volto code, fix alignment images, fix text "Updating the package code"

ksuess:


docs/mastering-plone/volto_richtexteditor.md
```diff
@@ -19,23 +19,7 @@ html_meta:
...
-
-```shell
-git checkout TODO tag to checkout
-```
-
-Code for the end of this chapter:
-
-```shell
-git checkout TODO tag to checkout
+Add a button and feature to the rich text editor DraftJS.
```
stevepiercy, line 22: Isn't this now Slate?

ksuess: The default editor since 16.0.0-alpha.15 is Slate, yes. I've written this chapter for the former editor DraftJS. I label it now with "DraftJS".
I think there are two options now:
- Adding a chapter on how to customize the editor Slate
- Leaving instructions on how to customize an editor to Volto documentation
Apart form this I would like to leave this DraftJS chapter in place for some weeks, as not everyone will switch instantaneous to the new default editor so this is still of interest.

stevepiercy: I think a new chapter on customizing Slate is necessary, no matter what. That's outside the scope of this PR.

Volto docs are changing. See the PR https://github.com/plone/volto/pull/3508 which covers multiple scenarios. We can ignore the upgrading scenarios, because that is outside the scope of Training as one-time deal. Folks might install a version of Volto that deprecates `draftJS`. As such, we need to include how to enable `draftJS` as the text block if we decide to maintain it going forward.

PR #638, opened 2022-10-03
Add patch scrollToActive (active main navigation item)

ksuess:
See issue https://github.com/plone/documentation/issues/1257


stevepiercy:
The patch works only when clicking an item in the left navigation, and not when clicking an item in a toctree or page content or visiting a link (https://deploy-preview-638--plone-training.netlify.app/advanced-python/data_model.html). It also does not scroll when clicking a nested item in the navigation. Is that the intent?

ksuess:
Please check out locally. Preview does not load js correctly.

stevepiercy (approved):

PR #644, opened 2022-10-07
Update Volto Hands-On

pbauer:


jackahl:
I reworked lots of the training to be more in line with what the fundementals on working with Volto are. The update is definetely not done but I think it should be in a state, to be usable during this years training at the conference. Smaller amendments will probably follow tomorrow.

sneridagh (approved):

PR #645, opened 2022-10-08
Update Training Docs for Classic UI theming from scratch in 2022

santonelli:


docs/theming/theme_scratch.md
```diff
@@ -174,24 +174,34 @@ http://localhost:8080/
...
+Navigate to the control panel and activate your add-on. After that step your site may look broken until we actually work on the new theme. Before we start theming we're going to add a copy of the main template to our theme package.
 
-You see some basic styling because a precompiled `src/plonetheme/tokyo/theme/css/theme.min.css` has been shipped with the template. Bevor we start theming we're going to add a copy of the main template to our theme package.
+```{note}
+You can always switch back to the default Barceloneta theme in the theming control panel: http://localhost:8080/Plone/@@theming-controlpanel
+```
 
 ## Override Main Template
 
 Copy the page template from `parts/omelette/Products/CMFPlone/browser/templates/main_template.pt` to `src/plonetheme/tokyo/browser/templates/main_template.pt`.
```
pbauer, line 197: I would use z3c.jbot to override the template. This way you do not need to change the zcml.

pbauer: Also I would mention they should make commit before modifying the template to have a clean diff.

pbauer (changes requested):

docs/theming/theme_scratch.md
MrTango: it does not work with jbot for the main_template as far as i know

docs/theming/theme_scratch.md
```diff
@@ -283,13 +301,17 @@ Let's make use of Bootstrap's layout system and add a `container`, a `row` and s
                     body_class python:plone_layout.bodyClass(template, view);"
         tal:attributes="class body_class;
                         dir python:isRTL and 'rtl' or 'ltr';
-                        python:plone_view.patterns_settings()"
+                        python:context.restrictedTraverse('@@plone_patterns_settings')();"
```
MrTango, line 304: i would keep this here in sync with the default main_template and provide plone_view, it's easier to use then.

MrTango (approved):

docs/theming/theme_scratch.md
pbauer: The changes that Stefan did are in sync with the current main_template

docs/theming/theme_scratch.md
pbauer: True, that's because of https://github.com/plone/Products.CMFPlone/issues/2666

santonelli: > Also I would mention they should make commit before modifying the template to have a clean diff.

I added it in the hints section.

docs/theming/theme_scratch.md
santonelli: I updated the template to get the current version.

pbauer (approved):

PR #655, opened 2022-10-10
naked domains must be backticked to prevent linkification

stevepiercy:


stevepiercy:
This PR fixes the failed build so that Training docs can be deployed again.

PR #680, opened 2022-11-01
Fix volto hands-on quickstart command to start the backend

davisagli:


stevepiercy (approved):
Yup. That image is not an environment variable.

Issue #688, opened 2022-11-07
Node 18 LTS not yet supported


stevepiercy:
[Training setup](https://training.plone.org/plone-deployment/setup.html#) was created just before Node 18 LTS released. Volto has not yet adapted to this version, so using LTS will fail.

```
Running sanity checks
  - Python: ✓
  - Node: Node version is not supported: Got v18.12.1
  - yo: ✓
  - Docker: ✓
  - git: ✓
================================================================================
ERROR: Stopping generation because pre_gen_project hook script didn't exit successfully
Hook script failed (exit status: 1)
```

The user must set:

```
nvm use 16
```

We should add a note stating the foregoing so that users do not get frustrated with an unsupported Node LTS version.

@jensens @ericof 

See also https://github.com/plone/training/issues/687

stevepiercy:
Fixed in https://github.com/plone/training/pull/690

PR #690, opened 2022-11-07
Add infos for node lts support

### LINKED ISSUE #687: Plone Deployment macOS tips
zcsh on macOS 10.15 requires that for nvm commands, its options with glob characters `*` be quoted.

https://training.plone.org/plone-deployment/setup.html#node-and-node-tools

```
nvm install lts/*
nvm alias default lts/*
```

Should be:

```
nvm install 'lts/*'
nvm alias default 'lts/*'
```

loechel:
a proposal to fix #687 and #688


jensens (approved):
Yes, this clarifies it.

loechel:
@stevepiercy feel free to merge, if you are ok with these changes

stevepiercy:
Thanks, @loechel, I pushed some English grammar fixes. This is very helpful. Merging after green.

ksuess:
Can we please next time make sphinx configuration changes in different PRs, @stevepiercy 
Changes like "Show TODO directives." could need some action for authors.

stevepiercy:
@ksuess `{todo}` directives they should be published. It shows that work needs to be done.  If we don't show them, it could mislead the user to think that everything is OK, and that would be much worse.

If instead the author intends a private comment, then they should use `% my-comment` or `% TODO: something for the author only` as that never shows and does not indicate future work is needed in a public way.

ksuess:
> @ksuess `{todo}` directives should be published. It shows that work needs to be done. If we don't show them, it could mislead the user to think that everything is OK, and that would be much worse.
> 
> If instead the author intends a private comment, then they should use `% my-comment` or `% TODO: something for the author only` as that never shows and does not indicate future work is needed in a public way.

A sphinx configuration change is off topic here. 
This change needs action and therefore should next time be visible as a separate PR.

PR #699, opened 2022-12-13
Fix links

pbauer:
Use 6.docs.plone.org and 5.docs.plone.org

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -221,8 +221,8 @@ html_theme_options = {
 # https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html
 #
 intersphinx_mapping = {
-    "plonedocs": ("https://docs.plone.org/", None),
-    "plone6docs": ("https://6.dev-docs.plone.org/", None),
+    "plonedocs": ("https://5.docs.plone.org/", None),
+    "plone6docs": ("https://6.docs.plone.org/", None),
     "python": ("https://docs.python.org/3/", None),
 }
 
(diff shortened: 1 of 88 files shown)


docs/conf.py
```diff
@@ -221,8 +221,8 @@
 # https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html
 #
 intersphinx_mapping = {
-    "plonedocs": ("https://docs.plone.org/", None),
-    "plone6docs": ("https://6.dev-docs.plone.org/", None),
+    "plonedocs": ("https://5.docs.plone.org/", None),
```
stevepiercy, line 224: I'd like to use an explicit version number going forward. This change will require a global find and replace. [My search turned up only this one instance](https://github.com/plone/training/search?q=plonedocs), so it should be trivial.

```suggestion
    "plone5docs": ("https://5.docs.plone.org/", None),
```

docs/deployment/index.md
```diff
@@ -57,5 +57,5 @@ letsencrypt-certbot
 ```
 
 ```{seealso}
-<https://docs.plone.org/manage/deploying/>
+<https://5.docs.plone.org/deployment/>
```
stevepiercy, line 60:
```suggestion
https://5.docs.plone.org/manage/deploying/
```

stevepiercy (changes requested):
This is good. Thank you!

- Needs a [global find for `6.dev-docs.plone.org` and replace with `6.docs.plone.org`](https://github.com/plone/training/search?q=6.dev-docs).
- Accept a couple fixes.

Then it is good to go.

Issue #727, opened 2023-03-07
[Training "React"] Updation needed in the folder structure of the React Training (bootstrap.html#bootstrapping-a-project)


rakshith6404:
The Folder structure which is given is to be updated in the react part of the training,the [Bootstrapping](https://training.plone.org/5/react/bootstrap.html#bootstrapping-a-project) part of the React training 

Folder structure given in the [Training](https://training.plone.org/5/react/bootstrap.html#bootstrapping-a-project) 

Current Structure when you make a React-app using ```npx create-react-app my-app```:

![Screenshot from 2023-03-07 14-47-54](https://user-images.githubusercontent.com/92712034/223381374-a4f2fa95-69f7-4bc6-bdf3-699be6360353.png)

stevepiercy:
Please submit a PR with a correction using the `tree` command. Thank you!

rakshith6404:
Closing this issue and marking this as done

PR #729, opened 2023-03-07
folder_structure changed in react-docs

rakshith6404:
I have recently created an issue on the Latest folder_structure regarding the issue https://github.com/plone/training/issues/727 and also have added using the ```tree``` command.

Looking forward for a MR :smile:


share/man/man1/ttx.1
```diff
@@ -0,0 +1,225 @@
+.Dd May 18, 2004
```
stevepiercy, line 1: Please remove this file.

stevepiercy (changes requested):

share/man/man1/ttx.1
rakshith6404: Workingo on it.

docs/react/bootstrap.md
```diff
@@ -50,15 +50,19 @@ my-app
 ├── public
 │   ├── favicon.ico
 │   ├── index.html
-│   └── manifest.json
+│   ├── manifest.json
+│   ├── robots.txt
+│   ├── logo512.png
```
stevepiercy, line 55: I don't think you ran `tree` in a shell, else these items would be sorted alphabetically.

Also `package-lock.json` is missing from the root of `my-app`.

stevepiercy (changes requested):

PR #730, opened 2023-03-07
updated the react folder_structure

rakshith6404:
I have created an issue on the Latest folder_structure regarding the issue https://github.com/plone/training/issues/727 and also have added using the tree command.

I created a new branch as i was getting conflict errors , Looking forward for a review for the same.

@stevepiercy , what if we could add a note saying and telling about them about there are many files and folders inside the node-modules folder?  

Screenshot from 2023-03-07 19-51-55

### DIFF
--- a/docs/react/bootstrap.md
+++ b/docs/react/bootstrap.md
@@ -44,6 +44,7 @@ It will create a folder called `my-app` inside the current folder with the follo
 ```console
 my-app
 ├── node_modules
+├── .gitignore
 ├── package.json
 ├── package-lock.json
 ├── public
@@ -63,6 +64,10 @@ my-app
     ├── logo.svg
     ├── reportWebVitals.js
     └── setupTests.js
+```
+
+```{note}
+We omit from the structure the files and directories inside the `node_modules` directory because there are too many to list.
 ```
 
 ## Running The Project
--- a/docs/react/bootstrap.md
+++ b/docs/react/bootstrap.md
@@ -43,8 +43,8 @@ It will create a folder called `my-app` inside the current folder with the follo
 
 ```console
 my-app
-├── node_modules
 ├── .gitignore
+├── node_modules
 ├── package.json
 ├── package-lock.json
 ├── public
--- a/docs/react/bootstrap.md
+++ b/docs/react/bootstrap.md
@@ -44,7 +44,6 @@ It will create a folder called `my-app` inside the current folder with the follo
 ```console
 my-app
 ├── .gitignore
-├── node_modules
 ├── package.json
 ├── package-lock.json
 ├── public
--- a/docs/react/bootstrap.md
+++ b/docs/react/bootstrap.md
@@ -43,14 +43,17 @@ It will create a folder called `my-app` inside the current folder with the follo
 
 ```console
 my-app
-├── README.md
 ├── node_modules
 ├── package.json
-├── .gitignore
+├── package-lock.json
 ├── public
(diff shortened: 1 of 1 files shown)


docs/react/bootstrap.md
```diff
@@ -43,22 +43,26 @@ It will create a folder called `my-app` inside the current folder with the follo
 
 ```console
 my-app
-├── README.md
 ├── node_modules
 ├── package.json
-├── .gitignore
```
stevepiercy, line 49: Please restore `.gitignore`. It might be invisible, but it definitely exists.

stevepiercy (changes requested):

stevepiercy:
> @stevepiercy , what if we could add a note saying and telling about them about there are many files and folders inside the node-modules folder?

Good idea. Suggest:

```
It will create a folder called `my-app` inside the current folder with the following structure.
We omit from the structure the files and directories inside the `node_modules` directory because there are too many to list.
```

rakshith6404:
I have updated the folder structure and also have added the note required which @stevepiercy have suggested.

docs/react/bootstrap.md
```diff
@@ -44,6 +44,7 @@ It will create a folder called `my-app` inside the current folder with the follo
 ```console
 my-app
 ├── node_modules
+├── .gitignore
```
stevepiercy, line 47: Fix sort order.

```suggestion
├── .gitignore
├── node_modules
```

stevepiercy (changes requested):

docs/react/bootstrap.md
rakshith6404: This was the order i got from when using ```tree``` command

docs/react/bootstrap.md
```diff
@@ -43,22 +43,31 @@ It will create a folder called `my-app` inside the current folder with the follo
 
 ```console
 my-app
-├── README.md
+├── .gitignore
 ├── node_modules
```
stevepiercy, line 47: Let's delete this, so that it aligns with the note.

```suggestion
```

docs/react/bootstrap.md
stevepiercy: That is... strange. What command do you use?

Try the following to go 2 levels deep, show all files including invisibles, and ignore the `node_modules` directory.

```
tree -L 2 -a -I node_modules
```

stevepiercy: In my output, `.` files are first, then files and directories that start with an uppercase letter, then files and directories that start with a lowercase letter.

stevepiercy (changes requested):

docs/react/bootstrap.md
rakshith6404: It was mistake that .gitignore was before the node_modules....i have updates the mistake...pls review it

rakshith6404:
This is what i get when i exclude the node_module folder using ```tree -I "node_modules\```": 
Screenshot from 2023-03-08 16-24-27

docs/react/bootstrap.md
rakshith6404: > Let's delete this, so that it aligns with the note.

Could you tell me where should the alignment be with the note?

stevepiercy (approved):

docs/react/bootstrap.md
stevepiercy: The note says we omit it. I committed the suggestion. This is good to merge. Thank you!

Issue #533, opened 2021-09-15
Add dark mode


stevepiercy:
Consider adding a dark mode for usability.

https://github.com/executablebooks/sphinx-book-theme/issues/265

stevepiercy:
Progress is being made on a [PR in the pydata-sphinx-theme](https://github.com/pydata/pydata-sphinx-theme/pull/540) on which sphinx-book-theme is based. If you have feedback, please let them know.

stevepiercy:
And now it is in progress in https://github.com/executablebooks/sphinx-book-theme/pull/574

stevepiercy:
This also includes the new announcement banner feature that is required in https://github.com/plone/training/issues/672

rakshith6404:
As i raised the issue on the same,i Would like to work on Add dark mode which i am currently going through the methods that were used in pydata and sphinx to create a dark mode.

stevepiercy:
@rakshith6404 please see https://6.docs.plone.org/contributing/#making-contributions-on-github, specifically Item 2. 😁

rakshith6404:
I have pretty much went through the contributing guidelines long back :smile:

stevepiercy:
Closed by #752

PR #752, opened 2023-04-02
Upgrade theme

stevepiercy:
Here are changes that I found. Some may affect our customizations, others I don't know if there is a replacement yet. We need to discuss.

- New search input removes our customized search input.
- Search results input loses CMD-K hint and UI is strange. I think we could modify the default by adding a clear button, instead of trying to preserve the existing UI.
- `html_theme_options.extra_navbar` is just gone. It's not even documented in any of the theme's ancestors. I don't know if we want to keep the stuff that got placed in the lower left. I am fine with letting it go. The fewer customizations, the easier it is to maintain.
- The title [Plone Training 2023 documentation](https://training.plone.org/#) just below the logo no longer appears. We still need this, I just don't know where it should go yet.
- Code blocks are illegible in dark mode.
- Scroll-spy is gone. This is the feature when you scroll, the subnavigation highlights the current section.

Closes https://github.com/plone/training/issues/533


stevepiercy:
### Update 2022-04-09

- Fixed requirements to pull in the latest versions of everything. We need a clean canvas, and can build up from there.
- Deleted a custom template `docs/_templates/sections/header-article.html` that appeared to do nothing. @ksuess or someone else might know its purpose, and if it is still needed, I can restore it.
- [Removed `extra_navbar` configuration as obsolete and redundant to new theme](https://github.com/plone/training/pull/752/commits/a0dd2413e8828771592e955bca7a653df943a5b1)
- Cleaned up a lot of broken links and references.
- Fixed broken HTML meta tag formats on a few pages. Note that I will not add missing attributes for the hundreds of warnings, as that is the responsibility of the author who cares about SEO for their training. I would encourage authors to use HTML meta information to increase their business's exposure.

stevepiercy:
### Update 2023-04-10

> * New search input removes our customized search input.

Won't fix. I much prefer the cleaner UI, now that I have gotten used to it.

> * Search results input loses CMD-K hint and UI is strange. I think we could modify the default by adding a clear button, instead of trying to preserve the existing UI.

Won't fix. CMD-K selects the input content and a tap of the DEL key is sufficient.

> * `html_theme_options.extra_navbar` is just gone. It's not even documented in any of the theme's ancestors. I don't know if we want to keep the stuff that got placed in the lower left. I am fine with letting it go. The fewer customizations, the easier it is to maintain.

We could add extra content to the footer. It's not required. I think a better option would be to think about how to style plone.org together with docs and training.

> * The title [Plone Training 2023 documentation](https://training.plone.org/#) just below the logo no longer appears. We still need this, I just don't know where it should go yet.

Fixed.

> * Code blocks are illegible in dark mode.

Fixed. Gutted `custom.css`.

> * Scroll-spy is gone. This is the feature when you scroll, the subnavigation highlights the current section.

Fixed.

- [ ] The search filter by training name needs to be restored.
- [ ] Need to switch to Sphinx Design for showing and hiding answers.

stevepiercy:
It's all ready to review. w00t!

There is a slight bug, to set the default value "all trainings" for the filter. I'll fix that.

stevepiercy:
Slight bug is now fixed.

polyester (approved):

fredvd:
Only issue with the theme I could fine:  viewing the trainings on tablet size  the hamburger menu to remove the left table of contents doesn't work in portrait mode.  (the pane with the plone logo at the top). If I hold my table horizontal, the folding of the pane does work. :-(  

dark mode  ❤️

fredvd (approved):

stevepiercy:
@fredvd can you replicate the issue on the site for the theme? https://sphinx-book-theme.readthedocs.io/en/stable/ If yes, then would you please open an issue with that theme? I think there is a range of breakpoints (952px-984px) where it does not work, but does work outside that range.

stevepiercy:
Actually I have opened an issue https://github.com/executablebooks/sphinx-book-theme/issues/721. Thank you @fredvd

fredvd:
> Actually I have opened an issue [executablebooks/sphinx-book-theme#721](https://github.com/executablebooks/sphinx-book-theme/issues/721).

Yes, it happens on the offical theme page as well. Thanks for opening the issue there. Hadn't tested yet on desktop myself.

PR #761, opened 2023-05-10
Replace spellcheck with Vale

### LINKED ISSUE #751: Use Vale for spell and style checking
We got it working in docs and Volto, let's do it here.

stevepiercy:
- Use latest versions in GitHub Actions
- Allow external branches to be built
- Add missing `make livehtml` command in setup-build.md
- Add section in .gitignore for Generated files from Vale and others

Closes #751


stevepiercy:
We can safely ignore the build error from Vale. I opened an issue and a PR to fix it.

- https://github.com/errata-ai/Microsoft/issues/51
- https://github.com/errata-ai/Microsoft/pull/52

stevepiercy:
My PR to Vale was merged. This is now ready for final review.

fredvd (approved):
LGTM.  Python 3.10 is now used, could we use 3.11 as the latest/greatest as well? Can also be bumped later if this works for now.  

In general I would advise in the local setup instructions to prefer latest available Python version as well when we update the docs. It now says minimum Python 3.8 but that gives a big spread of possible edge cases in the years to come in the stack.  Python3.8 goes out of asupport autumn next year, which will be some maintanance again then to bump the minimal requirement.

stevepiercy:
This repo's requirements should align with those for the [current version of Plone](https://6.docs.plone.org/install/install-from-packages.html#pre-requisites-for-installation). That's easier to remember or lookup. When requirements are added or dropped in Plone, we should do the same in Training and Documentation.

PR #781, opened 2023-08-15
Bring back breadcrumbs to search results

ksuess:


stevepiercy (commented):
If either Sphinx, or the parent or grandparent themes, are updated in the future, will we need to update this again? I don't see an obvious way to avoid it. Perhaps switching to Nuclia for search results and disabling this would be the way? Or, along those lines, could we create a `custom-search-tools.js` and specify that in our `conf.py` as a replacement for the default `searchtools.js`?

The main reason I am thinking about this is to create a Python package for our [customized Sphinx theme](https://github.com/plone/plone-sphinx-theme) (I haven't pushed to it yet), which we can then install. We did that for all the projects under the Pylons Project, and that worked very well.

stevepiercy:
@ksuess also thank you very much for restoring the breadcrumbs. @justdaksh this might be of interest to you for the Nuclia search GSoC project.

ksuess:
> If either Sphinx, or the parent or grandparent themes, are updated in the future, will we need to update this again? I don't see an obvious way to avoid it. Perhaps switching to Nuclia for search results and disabling this would be the way? Or, along those lines, could we create a `custom-search-tools.js` and specify that in our `conf.py` as a replacement for the default `searchtools.js`?
> 
> The main reason I am thinking about this is to create a Python package for our [customized Sphinx theme](https://github.com/plone/plone-sphinx-theme) (I haven't pushed to it yet), which we can then install. We did that for all the projects under the Pylons Project, and that worked very well.

Yes, it is a customization, which needs to be checked if updates are needed on a Sphinx / Sphinx book theme upgrade. This is not ideal and I would be happy to avoid that.
I think if the Nuclia search is ready and does provide what we need, in my opinion breadcrumbs in search results are a need, then this customizations can be removed. 
I am on the way to update the mastering plone development training. And the search does help a lot with fifty chapters.

And, yes ,a "Plone documentation and training" Sphinx theme would be great.

PR #792, opened 2023-09-03
Integrating Nuclia Widget

justdaksh:
CC: @stevepiercy 
Here is a draft for Nuclia widget integration into training.

Some of the issues to be solved:

- [ ] Null and Undefined results 
- [ ] Loading of widget w.r.t page and Breadcrumbs w.r.t results in Offset.


docs/_static/searchtester.js
```diff
@@ -0,0 +1,143 @@
...
+    }
+    return breadcrumbs;
+  } catch (error) {
+    console.error("Error fetching mapping:", error);
+    return [];
+  }
+}
+
+// Function to add breadcrumbs above each result heading
+function addBreadcrumbsToResults() {
```
ebrehault, line 34: Actually, I do not have a better approach than that for now.
Working on the DOM directly like this is far from ideal, it is difficult to maintain and might not be stable over time (typically, there is guarantee that the CSS selector using the `svelte-1yttzcg` class will work for ever, this classname is generated automatically by the Svelte compiler).
I need to think about how I could make the Nuclia Widget more customisable.

upload.py
```diff
@@ -0,0 +1,104 @@
...
+    title = extract_first_heading(path)
+    origin_url = f"{PUBLIC_URL}{path.replace('.md', '').replace('./docs/', '')}"  # noqa
+    sdk.NucliaUpload().text(
+        path=path,
+        format="MARKDOWN",
+        slug=slug,
+        title=title,
+        url=KB,
+        api_key=API_KEY,
+        origin={"url": origin_url},
```
ebrehault, line 59: you could get the breadcrumbs data for the current doc and add it here:
```
extra={"metadata": {"breadcrumbs": [
    {"url": "https://training.plone.org/", "label": "Plone Training"},
    {"url": "https://training.plone.org/contributing/index", "label": "Contributing to Plone Training"},
]}}```

File: docs/_static/searchtester.js [hunk 3]
```diff
@@ -0,0 +1,156 @@
+const { switchMap } = rxjs;
+const nucliaResult = document.querySelector("nuclia-search-results");
+const shadowRoot = nucliaResult.shadowRoot;
+
+const nuclia = new window.NucliaSDK.Nuclia({
+  backend: "https://nuclia.cloud/api",
+  zone: "europe-1",
+  knowledgeBox: "62407006-2711-4631-9c03-761d156de289",
+});
```
justdaksh, line 9: I accidentally pushed the kb token here 
I think we should make another secret for `knowledgeBox`

ebrehault: @justdaksh Don't worry, that's not a token, that's the kb id, and that's not really a secret because the kb id is in the URL of the requests made by the widget, so it is necessarily publicly visible.

stevepiercy:
Here's that git command from Volto that builds only when docs change:

https://github.com/plone/volto/blob/master/netlify.toml#L5

docs/_templates/search.html
```diff
@@ -1,42 +1,79 @@
...
+<script>
+  // Activate the search field on page load 
+  let searchInput = document.querySelector("form.bd-search input");
+  if (searchInput) {
+    searchInput.focus();
+    searchInput.select();
+    console.log("[PST]: Set focus on search field.");
+  }
+</script>
+<!-- I Dont think below code is required anymore as it is related to sphinx-->
```
stevepiercy, line 69: I partially agree, but the whitespace changes make it difficult to comment. Please revert the whitespace changes, and `@` me when I can review again. I think that the `<script>` tags for `searchtools.js` and `searchindex.js` may be removed from the template. I am not sure about `language_data.js`.

requirements.txt
```diff
@@ -13,3 +13,5 @@ sphinx-sitemap
 sphinx-togglebutton
 sphinxcontrib-spelling
 sphinxext-opengraph
+nuclia
+nucliadb-sdk
```
stevepiercy, line 17: Please sort these two new entries into the preceding requirements

stevepiercy (changes requested):
First round of review, mostly just formatting and style.

stevepiercy:
I have not yet commented on design of the Nuclia widget, as I was waiting for some dust to settle. Better late than never. In general it would be good to reproduce as close as practical the existing search results listing, as @ksuess put significant thought and effort into its style and usability: https://training.plone.org/search.html?q=schema

1. I think that the icon to the left of each listing provides nothing useful, especially because we are indexing only text. All results are text, so it feels redundant. Can it be removed?
2. In the breadcrumb, only the first part—the training name—should be hyperlinked, and that link should be to the index of that training.
3. The name of the page and subheadings is hyperlinked, but it should be blue instead of white.
4. Hyperlinks should be `--pst-color-primary: #579aca;` ("pst" for PyData Sphinx Theme, the parent theme)
5. I prefer sufficient whitespace over a horizontal line between each listing. Less clutter. After the above changes are made, it's possible nothing needs to be done for this item. We'll see.

If there are other things that you see, let's chat first. These are the obvious things that caught my eye.

I really like the infinite scroll loading of results as needed, and "chat with your docs" feature.

justdaksh:
cc: @stevepiercy @ebrehault 

**I made some changes locally and here is how the results look now:**

> image

> image

==================================================================================
**Changes:**

1.  Nuclia Widget provides an option to hide thumbnails, now it forms a small icon on the left. Do we want to remove it too?
2.  Do we want only the first part of every breadcrumb, `Plone Training`, hyperlinked and the remaining in non-hyperlinked text as our search has every breadcrumb starting from `Plone Training`. Except for the last item, every other item is hyperlinked.
3.  Headings and Subheadings now are `--pst-color-primary: #579aca;`. 
4.  Now our breadcrumbs, headings, and subheadings are of the same color.
5.  `requirements.txt` items have been sorted and Whitespaces have been reverted back in `search.html`.

**Issues:**

- [ ] Null results
- [ ] Breadcrumbs and widget loading speed. 
- [ ] Breadcrumbs Only update when you have scrolled down and results have been updated twice on scroll.

I am still working to find a better logic for the above two issues to work out.

**Additional:**

1. Color change on hover on breadcrumb links from `--pst-color-primary: #579aca;` to `--color-on-hover: #0056b3;`. 
2.  Widget `mode` is now in sync with the `light/dark` mode of the site. 

Do we want any of these?

stevepiercy:
- I don't see the changes. Please push commits to this branch so I can preview them locally and make sensible comments.
- Toggle to light mode. You'll see. Even if you start in light mode and reload, the issue is the same. I don't know if there is a way to handle this in Nuclia's SDK. For this new issue report, because it comes up so late in the process, it would be unfair to make this a new goal. If you see something obvious, then great, go for it, but please do not spend a lot of time on it.

.github/workflows/nuclia_sync.yml
```diff
@@ -0,0 +1,42 @@
...
+        run: |
+          python3 breadmaker.py
+
+      - name: Commit changes
+        run: |
+          git config --global user.name github-actions
+          git config --global user.email github-actions@github.com
+          git add .
+          git commit -m "Nuclia Sync: Updated docs" --allow-empty
+          git push
```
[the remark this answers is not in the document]
justdaksh: @ebrehault I have put the check for `nuclia_sync.json` only. This should work, right?

stevepiercy:
Final items to work on:

- [ ] Increase search item's title font size.
- [ ] Increase space between each search result item.
- [ ] Keep width of search results to 896px.

stevepiercy:
This is much easier to read. Thank you!

I found a couple issues after `make livehtml` and previewing locally. 

- [ ] The title link needs to have a hover state. An underline would be sufficient, and is typical. Here's what PyData theme does:
  ```css
  a:hover {
    text-decoration-skip: none;
    color: var(--pst-color-link-hover);
    text-decoration-skip-ink: none;
    text-decoration-thickness: max(3px,0.1875rem,0.12em);
  }
  ```
- [ ] In dark mode, search for `schema`, then hover over the "Display all results" link. It turns black. This should be readable. I'd suggest a grey that has sufficient contrast in both light and dark mode. Screenshot 2023-10-01 at 6 22 32 PM
- [ ] When clicking a search result, it scrolls to the location of that result in the page, and highlights it with pink, which is a reserved color for inline code. Here's what the PyData theme uses:
  ```css
  dt:target, span.highlighted {
    background-color: var(--pst-color-target);
  }
  ```

stevepiercy:
I've just updated this branch. I tried to get it to work with Sphinx 8.1.2, but had no luck in local development. I have no idea what to do to get it to work. @justdaksh I know it's been a long time, and I understand if you've moved on. Is this something you would like to work on? Also pinging @ebrehault to discuss.

ebrehault:
Hi @stevepiercy !
I see we have errors in the sync script (probably a minor issue), but what is the problem you get with Sphinx exactly?

stevepiercy:
@ebrehault I updated again with the Plone Sphinx Theme and recent merges to `main`, falling back to Sphinx 7.4.7. The Nuclia search widget does not appear.

I could post the JavaScript console with its 10 errors, but nothing is obvious to me, so I reckon it goes deeper.

ebrehault:
@stevepiercy I have pushed a fix, the widget is not displayed.
I have also updated the options of the Nuclia widget.

stevepiercy:
@ebrehault thanks for the assist. It works for me now.

I restarted styling things, too.

Looking at the search results, I see a lot of MyST markup in the results. This is not good. I was checking out the docs for a better option, and found [Index a full website](https://docs.nuclia.dev/docs/ingestion/how-to/index-website). It looks like I can tell Nuclia to index only the content in the CSS selector `article.bd-article section`. Before I go that route, which method yields better search results, indexing MyST or HTML?

If it turns out to be better to index HTML, then we can add a CI step to build the site first.

Please let me know your thoughts. Thank you!

ebrehault:
Indexing MyST or HTML will make no difference in terms of quality of the search results (unless the HTML contains side contents like the navigation menu, the footer, etc., that might produce unwanted matches).

We have 3 possibilities:

- index the MyST contents (as of now), and if we want to avoid MyST markup in the results, we would need to convert it to pure Markdown (not sure if that's possible)
- index HTML contents after the build, and to avoid unwanted side content, we will have to parse the files with BeautifulSoup or equivalent to get the core content
- index the online web pages after deployment, in that case the Nuclia API lets us define a CSS selector to restrict the indexation to the core content).

stevepiercy:
- MyST contains both reStructuredText and Commonmark+, and there is no reliable converter to plain old Markdown. This is probably not the best choice.
- Build HTML, then index, sounds good. It sounds like Nuclia does not support directly indexing HTML from files that are not hosted on a web server, and that an intermediary parser is required. Is that correct? If so, then BS is fine to scrape the content from `<article.bd-article><section ...>`, and I can figure out the details of generating a file that Nuclia can consume.
- Index online is too late, but sounds like the easiest option.

ebrehault:
Nuclia does process local HTML files, but then we do not have the CSS selector option to narrow down to a part of the page, so we need to do that manually.

justdaksh:
hey @stevepiercy, apologies for the delay I was out of town for two weeks and will be back home the day after tomorrow. I'll get to this as soon as I reach back.

stevepiercy:
There's an error in the CI: https://github.com/plone/training/actions/runs/11968576684/job/33367643251?pr=792

ebrehault:
@stevepiercy my guess would be the `DEPLOY_NUCLIA_URL` is not set.

ebrehault:
@stevepiercy yes that was missing.
I have added the `DEPLOY_NUCLIA_URL` and `DEPLOY_NUCLIA_TOKEN` variables in this repo, it should work now.

stevepiercy:
Thanks @ebrehault. I was looking for them in https://github.com/plone/training/settings/variables/actions, and `DEPLOY_NUCLIA_URL` just showed up, but not `DEPLOY_NUCLIA_TOKEN`.

stevepiercy:
Oops, there it is! nm.

stevepiercy:
I have no clue what this failure message means.

```
Run CHANGED_FILES=$(git diff --name-only c0e95e9ffc38cf3450ec7a39caedaa68f7c8ecd7 b89c046e8bd1c5df4345297bc8ed35708bbc7c9b)
  CHANGED_FILES=$(git diff --name-only c0e95e9ffc38cf3450ec7a39caedaa68f7c8ecd7 b89c046e8bd1c5df4345297bc8ed35708bbc7c9b)
  echo "Changed files: $CHANGED_FILES"
  if echo "$CHANGED_FILES" | grep -q 'nuclia_sync.json'; then
    echo "::set-output name=changed::true"
  else
    echo "::set-output name=changed::false"
  fi
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.10.15/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.10.15/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.15/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.15/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.15/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.10.15/x64/lib
fatal: bad object c0e95e9ffc38cf3450ec7a39caedaa68f7c8ecd7
Error: Process completed with exit code 128.
```

Where do we get the git hash?

See: https://stackoverflow.com/a/11834472/2214933

stevepiercy:
Answering my own question.

```yml
          CHANGED_FILES=$(git diff --name-only ${{ github.event.before }} ${{ github.sha }})
```

Next question, are these the correct values to use? It rejected `github.event.before`.

stevepiercy:
@justdaksh where did you get `github.event.before`? I can't find it anywhere in the [GitHub docs for events](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows).

justdaksh:
I read about it from StackOverflow at that time, but I can't find the reference to it right now.

stevepiercy:
This one?

https://stackoverflow.com/a/61861763/2214933

stevepiercy:
From a comment in that SO post, I found this:

https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4:

> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set `fetch-depth: 0` to fetch all history for all branches and tags. Refer [here](https://docs.github.com/actions/using-workflows/events-that-trigger-workflows) to learn which commit `$GITHUB_SHA` points to for different events.

Commit incoming. We'll see.

stevepiercy:
That seems to have done the trick.

ksuess:
This looks promising and is well advanced. When can we get it finished and installed?

I played with the search.
Here is the search for "register panel". It gives some overview and the lists the chapter I expect in first place of the search results. That's fine.
https://plone-training--792.org.readthedocs.build/search.html?q=register+panel

Feedback

- [ ] Breadcrumbs with category (backend, frontend,…) would be helpful
- [ ] Remove myst formatting stuff from search results

stevepiercy:
> When can we get it finished and installed?

It's one of my highest priorities for Plone docs, but I have three Plone projects ahead of this item. The release of Plone Sphinx Theme 2.0.0 is one of them and is a partial blocker to this issue, since it will remove Sphinx Book Theme from the dependency stack. That said, I don't think the removal of Sphinx Book Theme will have a significant impact, except possibly some CSS rules.

I'll be busy through the end of March. I'd welcome any help to see this through.

PR #793, opened 2023-09-09
Effective Volto 2023

sneridagh:


docs/effective-volto/addons/blockdataform.md
```diff
@@ -112,3 +115,59 @@ a Select control for the active variation, in the sidebar.
 Note: you can assign the schema to `blockSchema` in the block configuration. It
 is used only to extract the block default values. This integration will be
 improved, in the future.
+
+## Using a dataAdapter pattern
+
+Sometimes is useful to adapt the incoming data to other data format, structure or type.
+You can use the dataAdapter pattern in `BlockDataForm` as this:
```
tiberiuichim, line 122: This pattern will cause confusion. 

Why is the BlockDataForm looking for a dataAdapter for a certain block? (it's not explained that the BlockDataForm is used internally by the teaser block).

It's not explained how to register the data adapter (although this would be obvious, in theory). The data adapter function uses CamelCase naming, usually reserved for components.

I think it's dangerous to have the data adapter directly call `onChangeBlock`. I think it's better if the purity of the data operations is kept. No side-effects. If the data adapter is called too late in the data "fixing" chain, then we need to figure out a solution to call it sooner, at the top of the chain, at the level of the "big form".

sneridagh: @tiberiuichim I completed the docs to make them more consistent and understandable.

Regarding the purity... we are already calling it virtually everywhere, since the default pattern is:

```js
          onChangeField={(id, value) => {
            onChangeBlock(block, {
              ...data,
              [id]: value,
            });
          }}
```

in all places. So calling it inside the dataAdapter is "legit", right? Granted, it could be handled better, but as a general improvement of the `BlockDataForm`.

docs/effective-volto/architecture/redux.md
```diff
@@ -82,3 +83,20 @@ const FormattedDate = ({
...
+different render trees, similar to Redux.
+
+```
+import { atom, useAtom } from 'jotai'
+
+const mySharedStateAtom = atom();
+
+function MyComponent(props) {
+  const [myVal, setMyVal] = useAtom(mySharedStateAtom);
+  return <button onClick={() => setMyVal(val + 1)}>Count up</button>
```
sneridagh, line 100: ❤️

PR #816, opened 2023-10-19
Fix mentioned filename extension

gforcada:
I'm going trough the `Plone 6 Classic UI Theming` and found this minor glitch 😄 

A general question though: which source is _leading_ regarding Classic UI theming? training.plone.org or docs.plone.org? 🤔 

Given that on the last Plone Conference this train was not given, I'm not sure what's more recent/maintained.

Thanks for all the work, I've gone through all points until here and I can only say that it's impressive the work on `plonecli`, `bobtemplates.plone`, `mrbob` and of course all the great work on Plone 6 🎉 🥳 🤩

### DIFF
--- a/docs/theming/theme_diazo.md
+++ b/docs/theming/theme_diazo.md
@@ -707,9 +707,9 @@ For example if you install a JQuery library you will find all JavaScript and CSS
 
 ## Adjust the layout and manifest files
 
-We need to adjust a bit the `manifest.ini`, to reflect the layout structure of the Bootstrap theme we got.
+We need to adjust a bit the `manifest.cfg`, to reflect the layout structure of the Bootstrap theme we got.
 
-By default the `manifest.ini` will look like this:
+By default the `manifest.cfg` will look like this:
 
 ```ini
 [theme]


petschki:
*I definitely would like to give some Classic-UI training in Brasília*

/cc @thet 😄

petschki:
see updated https://6.docs.plone.org/classic-ui/theming/barceloneta.html#theme-structure

stevepiercy:
Thank you!

gforcada:
Thanks! 🌟

MrTango (approved):

MrTango:
@gforcada the training is not really outdated. maybe in terms of bootstrap version and i think the dark mode is not in there, but the rest is up to date with Plone 6 from 2022 conference.
Thanks for the appreciation on the plonecli work. I'm always happy to see people using it and giving feedback. it probably need some updates for the addon template towards plone meta aso. maybe you can help there.
In the docs we try to keep everything short and as a reference, the docs should not turn into a copy of the trainings. if somebody doesn't know much about plone dev, the training is the first place to start. but if somebody knows a bit or is under guidens, she should be pointed into the section of the docs for a specific topic. like build a content type or a view. no need to walk through a one week training. the docs are the right place then.

PR #820, opened 2023-10-24
Fix const name in export default

rber474:
Name was misstyped having a lowercase instead an uppercase


stevepiercy:
@jackahl @rber474 I changed the base to `2024`, which is the last year this training was published. I think it was last presented in 2022, and it's now obsolete, but I'll merge this anyway for history.

PR #834, opened 2024-01-23
 Including essential details for beginners.

Vivek-04022001:
I've discussed this issue on the Plone community forum: https://community.plone.org/t/news-item-not-appearing-on-plone-news-page/18625?u=vivek-04022001. As a result, I've decided to submit a pull request to address the problem.


ksuess:
I understand that you are puzzled that the news items do not show up instantaneously after creating them. But this is not the right place to create a listing block. Showing contained items of a folderish item with a listing block is explained some lines later in https://training.plone.org/mastering-plone/features.html#folderish-content
Nonetheless this chapter should be enhanced in the flow. Will consider your suggestion in the upcoming overhaul. Thanks for the feedback!

PR #838, opened 2024-02-18
Update dexterity_reference.md

tjprescott:
As written, this triggers a crash. 
`AttributeError: 'datetime.date' object has no attribute 'date'`

Since `value` is already a `datetime.date` object, you can't call `.date()` on it.


ksuess (commented):
Your change would be necessary for a Date field, but not for a DateTime field as in the example of the tutorial.

PR #841, opened 2024-03-14
Workflows bump versions, remove duplicate jobs

stevepiercy:


requirements.txt
```diff
@@ -12,3 +12,4 @@ sphinx-sitemap
 sphinx-togglebutton
 sphinxcontrib-spelling
 sphinxext-opengraph
+vale==2.30.0
```
ksuess, line 15: Maybe we better follow the breaking change of the new file structure for the vale configuration, instead of pinning a former major vale version. For sure you have already read https://github.com/errata-ai/vale/releases/tag/v3.0.0

ksuess (changes requested):

requirements.txt
stevepiercy: Yes, I have an open issue in https://github.com/plone/documentation/issues/1602, but that is outside the scope of this PR. This is a temporary fix, until I have time to solve it properly.

ksuess: Ah, I see. Thanks for the hint.

ksuess (approved):

PR #842, opened 2024-03-21
Fix irregular plural noun

agan-k:
In the chapter plural of "criterion" is written as both "criteria" (correct, more common) and "criterions" (less common and possibly incorrect). In either case, using two different ways to express the same meaning makes it confusing.

### DIFF
--- a/docs/mastering-plone/volto_frontpage.md
+++ b/docs/mastering-plone/volto_frontpage.md
@@ -57,7 +57,7 @@ To understand why we need a collection criterion for a dynamic frontpage in Volt
 :align: left
 ```
 
-In the sidebar we see the `criteria` selection and if we click there, it'll show some of the selectable criteria ordered in categories like the following:
+In the sidebar, we see the {guilabel}`Criteria` select menu, and if we click there, it'll show some of the selectable criteria ordered in categories like the following:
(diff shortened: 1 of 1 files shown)


docs/mastering-plone/volto_frontpage.md
```diff
@@ -57,14 +57,14 @@ To understand why we need a collection criterion for a dynamic frontpage in Volt
 :align: left
 ```
 
-In the sidebar we see the `criteria` selection and if we click there, it'll show some of the selectable criterions ordered in categories like the following:
+In the sidebar we see the `criteria` selection and if we click there, it'll show some of the selectable criteria ordered in categories like the following:
```
stevepiercy, line 60: For UI bits and pieces, we should use semantic markup with `guilabel`, instead of inline literal markup which is usually used for inline commands or code.

```suggestion
In the sidebar, we see the {guilabel}`Criteria` select menu, and if we click there, it'll show some of the selectable criteria ordered in categories like the following:
```

stevepiercy (changes requested):
Thank you for the report. I made another suggestion to consider.

stevepiercy (approved):
Thank you!

PR #856, opened 2024-10-10
Fix broken links, set linkcheck_rate_limit_timeout=10

stevepiercy:


stevepiercy:
@ksuess this is the first of three PRs to bring the build stack up to date for Training.

There is a new configuration for `linkcheck_rate_limit_timeout = 10.0` which is ignored as an `int` so needs to be set as a `float` for Sphinx<=7.3.7. Even then, it stalls on  https://github.com/search, so I just ignored that URL. Anyway, now linkcheck is back to taking about 5 minutes to complete, instead of over 90 minutes.

PR #864, opened 2024-11-09
Move Python virtual environment into venv from root

stevepiercy:
`--clear` is the recommended way to wipe and recreate a clean Python virtual environment. `python3 -m venv venv --clear` will purge files and directories, such as `.git` and `.my_special_config`, from the current directory. This would be Very Bad™.

- Enhance `make distclean` such that it cleans both the docs _build directory and the virtual environment, then recreates the virtual environment and installs requirements into it. This is useful when your project gets confused about what exactly is in your virtual environment.
- Update internal Makefile command. No change to user commands.

Credit to @mrtango

----
📚 Documentation preview 📚: https://plone6--864.org.readthedocs.build/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,8 +191,8 @@ linkcheckbroken: venv/bin/python  ## Run linkcheck and show only broken links
 
 .PHONY: vale
 vale: venv/bin/python  ## Run Vale style, grammar, and spell checks
-	bin/vale sync
-	bin/vale --no-wrap $(VALEOPTS) $(VALEFILES)
+	venv/bin/vale sync
+	venv/bin/vale --no-wrap $(VALEOPTS) $(VALEFILES)
 	@echo
 	@echo "Vale is finished; look for any errors in the above output."
 
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    = 
-SPHINXBUILD     = "$(realpath bin/sphinx-build)"
-SPHINXAUTOBUILD = "$(realpath bin/sphinx-autobuild)"
+SPHINXBUILD     = "$(realpath venv/bin/sphinx-build)"
+SPHINXAUTOBUILD = "$(realpath venv/bin/sphinx-autobuild)"
 PAPER         =
 DOCS_DIR      = ./docs/
 BUILDDIR      = ../_build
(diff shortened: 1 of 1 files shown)


ksuess (approved):

Makefile
```diff
@@ -29,76 +29,80 @@ clean:  ## Clean build directory
...
 .PHONY: distclean
-distclean:  ## Clean docs build directory and Python virtual environment
+distclean:  ## Clean docs build directory and Python virtual environment, then install requirements
 	cd $(DOCS_DIR) && rm -rf $(BUILDDIR)/
-	rm -rf ./bin/ ./lib/ ./lib64 ./include ./pyvenv.cfg
+	python3 -m venv venv --clear
+	venv/bin/python -m pip install --upgrade pip
+	venv/bin/pip install -r requirements.txt
+	@echo
+	@echo "Installation of requirements completed."
```
ksuess, line 38: I like that it does not only clear, but also set up all fresh.

davisagli: This is a bit surprising in comparison to other Makefiles. I'm used to having one command to remove everything and another command to install everything.

And I don't think it's really needed. If `make distclean` didn't do this, things would still get installed the next time the user runs a command like `make html`, because it depends on `venv/bin/python`.

I don't feel strongly about this. Just reporting my reaction.

stevepiercy: By using `python3 -m venv venv --clear` alone, that blasts away the venv and recreates it _without requirements,_ so the next time `make html` is invoked, the requirements to _do not get installed_. It threw me for a loop until I figured out why `make html` stopped working after `make distclean`. You can try it out and see if you get the same results I did. Maybe I goofed somewhere.

davisagli: @stevepiercy I see. That could be fixed by replacing the `venv/bin/python` requirement with a file that is only there once the requirements have been installed (such as venv/bin/sphinx-build)

Basically you've got a bug in the Makefile where there are a lot of commands that depend on sphinx-build being installed, but they only declare that they depend on venv/bin/python existing, and there's a way for that to happen without installing sphinx-build.

stevepiercy: @davisagli so change this:

```make
venv/bin/python:  ## Setup up Python ...

...
.PHONY: deps
deps: venv/bin/python docs/volto docs/plone.restapi docs/plone.api  ## Create Python virtual environment, ...
```

to this?

```make
venv/bin/sphinx-build:  ## Setup up Python ...
...
.PHONY: deps
deps: venv/bin/sphinx-build docs/volto docs/plone.restapi docs/plone.api  ## Create Python virtual environment, ...
```

davisagli: @stevepiercy Yes -- but also the same change for every command which currently declares a dependency on venv/bin/python but actually runs $(SPHINXBUILD)

stevepiercy: That means I must use `rm -rf venv`. Oh, well, so much for `--clear`. PR incoming. I'm not going to rename `venv/bin/python` to `venv/bin/sphinx-build`. It doesn't really gain anything in the Makefile because it already installs requirements. PR #864

davisagli:
> It doesn't really gain anything in the Makefile because it already installs requirements.

Except in the situation we've been discussing, which is why I suggested it. But I'll move on, it's an edge case.

PR #865, opened 2024-11-10
Don't recreate the virtual environment, just delete it

stevepiercy:
See #864

----
📚 Documentation preview 📚: https://plone6--865.org.readthedocs.build/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,12 +30,8 @@ clean:  ## Clean build directory
 
 .PHONY: distclean
 distclean:  ## Clean docs build directory and Python virtual environment, then install requirements
+	rm -rf venv
 	cd $(DOCS_DIR) && rm -rf $(BUILDDIR)/
-	python3 -m venv venv --clear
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -30,12 +30,8 @@ clean:  ## Clean build directory
 
 .PHONY: distclean
 distclean:  ## Clean docs build directory and Python virtual environment, then install requirements
```
ksuess, line 32:
```suggestion
distclean:  ## Clean docs build directory and Python virtual environment
```

ksuess (changes requested):

davisagli (commented):
@stevepiercy If you don't want to fix the dependencies of the commands that require sphinx-build to be installed, then I think it's probably better for this to stay as is.

stevepiercy:
@davisagli yeah, I considered renaming, but it really didn't get us anything. If you have a `venv/bin/python` from `make html` or any other make command that installs it, then you also have the requirements.

PR #874, opened 2024-11-18
Fix rtd build 2

stevepiercy:
----
📚 Documentation preview 📚: https://plone6--874.org.readthedocs.build/


stevepiercy:
This is strange. The link that RTD generates and writes back to the description is incorrect. It should be https://plonetraining--874.org.readthedocs.build/. I'll merge this PR and try another, as perhaps the GH workflow needs to have the correct slug on `main` first.

PR #875, opened 2024-11-18
Fix spelling of Node.js

stevepiercy:
----
📚 Documentation preview 📚: https://plonetraining--875.org.readthedocs.build/


stevepiercy:
Yup, that was it. The value of `slug` on `main` in https://github.com/plone/training/blob/c03a2be519c53158298ea13a843a204bb120e5fd/.github/workflows/rtd-pr-preview.yml#L21-L21 is what matters.

@ksuess would you please double-check?

stevepiercy:
OK, one more cycle through, and it should work going forward. Basically we do not want `main` to be published or active, but we do want `legacy` as shown here:

https://plone-training.readthedocs.io/

PRs should now get built at `https://plone-training--PR#.org.readthedocs.build/`.

Merging to get this going, and to do one more final test.

PR #881, opened 2024-11-20
Shifting from .md to html injestion

justdaksh:
@stevepiercy Nuclia Knowledge Base has now been indexed with HTML

### Issues
- [ ] Some result's URL appear to be broken
- [ ] Bug in CI workflow 
- [ ] Breadcrumbs not visible in search results


.github/workflows/nuclia_sync.yml
```diff
@@ -1,46 +1,35 @@
...
     runs-on: ubuntu-latest
 
     steps:
       - name: Check out code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
-      - name: Set up Python
-        uses: actions/setup-python@v4
+      - name: Set up Python ${{ matrix.python-version }}
```
stevepiercy, line 14: Where is this declared? There is no `matrix` of Python versions. There is only one. I'd revert this line.

```suggestion
      - name: Set up Python

```

upload.py
```diff
@@ -15,87 +16,175 @@
...
 def normalize_path(path):
-    return path.replace('.md', '').replace('./docs', '').lstrip('/')
+    path = path.replace('./_build/html/', '')
+    return path.replace('.html', '').lstrip('/')
 
 
 def create_url(origin_url, url_path):
     return f"{origin_url.rstrip('/')}/{url_path}"
 
+def transform_path_to_url(path, base_url):
```
stevepiercy, line 88: I'm not sure we should do a transformation. Some trainings follow the `_` convention, others `-`. Is there a reason to do a transformation?

stevepiercy (commented):
A couple of questions. Otherwise this looks good.

justdaksh:
Ahh yes, `transform_path_to_url` function just merges `normalize` and `create URL`. I forgot to remove them but I have fixed that now.
when creating URL for `volto_customization` training we need `-` instead of `_` so the function just does that only.
so we get https://training.plone.org/volto-customization/index.html
instead of https://training.plone.org/volto_customization/index.html

stevepiercy:
Yes, but why do we get `volto_customization/index.html` in the first place? I refactored that to use `-` a few days ago.

justdaksh:
I guess the fix might not have been synced with this branch

stevepiercy:
Ah, yes, that would do it. Let me do that now, and we can take another look.

Also, one training will be added soon (I hope) and another will go away, so that synch will need to happen again.

stevepiercy:
@justdaksh done! Pull and let's have a look together.

justdaksh:
yay! URLs don't break anymore.

stevepiercy:
We still have CI failure, but let's handle that in the base branch PR in https://github.com/plone/training/pull/792#issuecomment-2493409144

stevepiercy:
CI passed in base PR, updated base branch here and resolved conflicts. We'll see if that resolves the CI issue.

stevepiercy:
CI passes. That leaves the breadcrumb issue.

Also I updated the RTD PR preview URL slug, so it is now at: https://plone-training--881.org.readthedocs.build/

PR #886, opened 2024-11-25
Update training to Plone Conference 2024

ericof:
----
📚 Documentation preview 📚: https://plone-training--886.org.readthedocs.build/


stevepiercy:
@ericof do you have the request/response diagram available as SVG or the source files you used to generate it? This would be a good Mermaid or Graphviz diagram for documentation.

ericof:
@stevepiercy Do we support Mermaid in our docs? I did not have time to check it, so I generated the png

stevepiercy:
@ericof yes, the recent release of sphinxcontrib-mermaid 1.0.0 now makes it work with Plone Sphinx Theme 1.0.0.

You can see some examples at https://6.docs.plone.org/conceptual-guides/package-dependencies.html. @jensens used https://www.mermaidchart.com/play with a live preview, then copy-pasted it into the docs, and I did a little fine tuning so that lines show up in dark mode. See also the docs for usage at https://6.docs.plone.org/contributing/documentation/myst-reference.html#diagrams-and-graphs-with-mermaid.

I think this diagram would be good on a new page under https://6.docs.plone.org/conceptual-guides/index.html, as it would describe the concept of the request-reponse cycle in Plone.

PR #908, opened 2024-12-19
Clarify content rules in the training

ghnire:
Updated the content rules section to be more explicit of how to create and apply a content rule.

----
📚 Documentation preview 📚: https://plone-training--908.org.readthedocs.build/


ksuess:
awesome. thank you @ghnire !

PR #918, opened 2025-01-27
Update deploy - add more explicit details to invoke and use `make test` command.md

acsr:
There is no hint for self students how to interpret the interactive output options of `make test` or how to detect missing results.

Also the switching of the context in terminal is not explicitly mentioned. So I suggest two more lines how to switch back and forth. The context can be guessed by the path of the makefile, but this can be overseen.

Remark: This is also my first test for the Quick & Dirty `Suggest Edit` workflow from the docs UI.

### DIFF
--- a/docs/plone-deployment/deploy.md
+++ b/docs/plone-deployment/deploy.md
@@ -26,22 +26,21 @@ You can find this stack at {file}`devops/stacks/ploneconf2024-<your-github-usern
 
 ## Building Docker Images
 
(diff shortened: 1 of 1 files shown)


stevepiercy (approved):

stevepiercy:
@acsr I restructured the bits and pieces to flow better.

This training is focused on deployment, not on how to interpret test results beyond pass or fail. So I removed the expected hints statement.

Please take a look, and let me know if it looks good to you. Thank you for contributing!

acsr:
Thanks for the overhaul and clarifying the scope of the test result mentioned. Go for it!

PR #919, opened 2025-01-27
Update deploy.md log Commands

acsr:
the log monitoring commans are not correct anymore. Montoring Logs Makefile commands in devops now have a `stack-` prefix and another ` make stack-logs-db` is available in setups created by cookieplone 0.8.2 using Plone 6.0.14, Volto 18.7.0.


stevepiercy (approved):
Thank you! Merging!

PR #924, opened 2025-02-24
Update troubleshoot.md | add hints for missing stack-create-site step

acsr:
see: https://community.plone.org/t/corepack-problems-specially-in-ci/20597

----
📚 Documentation preview 📚: https://plone-training--924.org.readthedocs.build/

### DIFF
--- a/docs/plone-deployment/troubleshoot.md
+++ b/docs/plone-deployment/troubleshoot.md
@@ -28,16 +28,24 @@ Under {menuselection}`Prefences --> Resources --> Advanced`, you can configure a
 -   [Change Docker Desktop settings](https://docs.docker.com/desktop/settings-and-maintenance/settings/)
 ```
 
+
 ## Docker memory usage
 
 Docker requires sufficient memory to install and build images.
 2GB of RAM is usually sufficient.
 See the previous section for settings details.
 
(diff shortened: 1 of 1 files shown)


acsr:
One last thing: The weird need for the pnpm workarounds shifted the focus on watching issues with the frontend containers caused by the pnpm issue. While testing over and over, the missing step was ignored. This is a common pattern in troubleshooting, when different vectors overlay each other and start misleading like a magicial confuses his audience to miss the point he uses to trick them.

stevepiercy (approved):
I tightened this up with English and MyST syntax, and pulling content from the [devops README](https://github.com/plone/cookieplone-templates/blob/main/project/%7B%7B%20cookiecutter.__folder_name%20%7D%7D/devops/README.md#plone-site-creation).

I'd like a Cookieplone Template technical review to ensure that this addition is correct and consistent. I'll request reviews from the appropriate individuals.

Thank you for bringing this up, @acsr!

acsr:
@stevepiercy 
> Thank you for bringing this up, @acsr!

Actually @fredvd figured it out finally.

davisagli (approved):
Thanks! It would be nice if we can also make the frontend report this error in a more helpful way, but the notes here are helpful in the meantime.

PR #928, opened 2025-02-28
Update index.md | Add an explicit Volto scope

acsr:
The title needs a clarification and we need a second topic pointing to the latest available training / docs fort Plone 6 Classic / similar to Plone 5

----
📚 Documentation preview 📚: https://plone-training--928.org.readthedocs.build/

### DIFF
--- a/docs/content-editing/index.md
+++ b/docs/content-editing/index.md
@@ -9,7 +9,7 @@ myst:
 
 (content-editing-label)=
 
-# Content Editing for Plone Volto
+# Content editing for Plone
 
 About
 : How to create and manage content using the Plone Volto frontend
--- a/docs/content-editing/index.md
+++ b/docs/content-editing/index.md
@@ -9,10 +9,10 @@ myst:
 
 (content-editing-label)=
 
-# Content Editing for Plone
+# Content Editing for Plone Volto
 
 About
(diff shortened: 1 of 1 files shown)


docs/content-editing/index.md
```diff
@@ -9,10 +9,10 @@ myst:
 
 (content-editing-label)=
 
-# Content Editing for Plone
+# Content Editing for Plone Volto
```
stevepiercy, line 12: There is no product called "Plone Volto". It's just "Plone". I also don't want this title to become verbose, as it is the only training for editing content in Plone.

However, headings should be sentence cased.

```suggestion
# Content editing for Plone
```

docs/content-editing/index.md
```diff
@@ -9,10 +9,10 @@ myst:
...
 
 About
-: How to create and manage content using Plone
+: How to create and manage content using the Plone Volto frontend
```
stevepiercy, line 15: I'll go along with more detail here, since it won't appear as a primary navigation item or a title of the training. But let's use appropriate terminology.

```suggestion
: How to create and manage content using Plone with Volto as its frontend
```

stevepiercy (changes requested):

PR #933, opened 2025-03-13
Add support for tooltips, `attrs_block`, `attrs_inline`, and `striket…

### LINKED ISSUE #540: Show tooltips in text for glossary terms and permalinks for definitions
From https://github.com/plone/training/issues/506#issue-937554502

There are several possible implementations.

- https://sphinx-hoverxref.readthedocs.io/en/latest/ (supported by the RTD team, but only works on sites hosted by RTD).
- https://docs.red-dove.com/sphinx_sizzle_theme/index.html#tooltips (supported only in the Sizzle theme).
- https://sphinx-panels.readthedocs.io/en/latest/index.html#link-buttons shows a couple of examples of tooltips on buttons and badges, but requires Bootstrap 4.

It appears that the downside of all these implementations is that they require the docs to reside on a server for the tooltip to make an XHR to grab the content from the reference. Else it's a lot of copy-paste for each instance of a `term`.

IMO this is too much work with not enough benefit, when users can open a link to a term in a new window or tab.

Instead I would suggest that we implement a custom style to indicate that a link is to a term in the Glossary. I think a dashed underline is typical for that.

stevepiercy:
…hrough`

- Add support for tooltips, `attrs_block`, `attrs_inline`, and `strikethrough`
- Closes #540

----
📚 Documentation preview 📚: https://plone-training--933.org.readthedocs.build/

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -62,17 +62,21 @@ extensions = [
     "sphinx_copybutton",
     "sphinx_design",
     "sphinx_sitemap",
+    "sphinx_tippy",
     "sphinxext.opengraph",
 ]
 
 # For more information see:
 # https://myst-parser.readthedocs.io/en/latest/syntax/optional.html
 myst_enable_extensions = [
-    "deflist",  # You will be able to utilise definition lists
-    # https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#definition-lists
-    "linkify",  # Identify “bare” web URLs and add hyperlinks.
-    "colon_fence",  # You can also use ::: delimiters to denote code fences,\
-    #  instead of ```.
+    "attrs_block",  # Support parsing of block attributes.
+    "attrs_inline",  # Support parsing of inline attributes.
+    "colon_fence",  # You can also use ::: delimiters to denote code fences, instead of ```.
+    "deflist",  # Support definition lists. https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#definition-lists
+    "html_image",  # For inline images. See https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#html-images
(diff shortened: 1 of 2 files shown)


stevepiercy:
@Manas-Kenge @ujsquared this PR adds features from your work in PST.

ksuess:
I thought Plone-sphinx-theme would be the place for sphinx extensions and their configuration. And that projects that use the PST theme would get these enhancements applied. If necessary they can overwrite.
What's the reason that here in training the sphinx extension tippy and the myst_enable_extensions configuration is updated, instead of just relying on an update of PST theme?

stevepiercy:
@ksuess PST is only a packaged theme that eliminates the need to copy over static assets, such as styles and JavaScripts that are included with the theme. However, it does not include configuration or its dependencies.

We need something like `plone/meta` for docs to update documentation configuration files. Without it, we have to copy-paste changes to requirements, `conf.py`, and other files. It's a maintenance burden.

The documentation cookiecutter template has this in mind, but it will only work on the initial setup, not ongoing maintenance.

Anyway, maybe we could start a new discussion in either `plone/meta`, `plone-sphinx-theme`, or the forum to create specifications? I'd be interested in doing some work on this.

acsr:
It make sense to add these new styling options to the MyST Reference in a second iteration. Please add a ticket related to that after approving the PR. I expect that there is nothing related.

stevepiercy:
@acsr these are maintained in both PST an documentation as the authoritative references. See https://github.com/plone/documentation/pull/1905.

acsr:
> @acsr these are maintained in both PST an documentation as the authoritative references. See [plone/documentation#1905](https://github.com/plone/documentation/pull/1905).

I missed to find a cross reference, sorry for the noise.

Issue #938, opened 2025-03-22
Issue regarding jsconfig.json on page /voltoaddons/01-addon-basics.html


codehashira99:
Please give appropriate instructions if our project has tsconfig.json file instead of jsconfig.json file , generated during project installation and setup.
The training documentation has instructions for jsconfig.json file, but if my project has already tsconfig.json , then should i delete tsconfig.json and create a new jsconfig.json?

PR #939, opened 2025-03-22
Addon is specified in package.json, not in project.json

GOGETASSB23:
To load an add-on, the project needs to specify the add-on in the addons key of package.json of main project.
 In the training  it was written project.json, I corrected it to package.json.

----
📚 Documentation preview 📚: https://plone-training--939.org.readthedocs.build/


davisagli (approved):

GOGETASSB23:
@stevepiercy sir, I think the #938 is regarding installation/update related.
#939 is a typo error , and they both are independent.
So, I think they can be addressed independently.

PR #944, opened 2025-04-07
Update Matomo tracking script

stevepiercy:
----
📚 Documentation preview 📚: https://plone-training--944.org.readthedocs.build/

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -3,10 +3,18 @@
 {{ super() }}
 <!-- Matomo Tag Manager -->
 <script>
-var _mtm = window._mtm = window._mtm || [];
-_mtm.push({'mtm.startTime': (new Date().getTime()), 'event': 'mtm.Start'});
-var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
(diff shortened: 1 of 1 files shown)


fredvd:
Note to self @fredvd :  need to check why tag container is used on 2023 and repeat for 2024 setup if required

stevepiercy:
@fredvd would you please review this PR? Do I need to revise it? Thank you!

fredvd:
Hi Steve, 

Sorry, totally missed this one.   Let me check this one. We switch from a mtm (matomo tag manager inclusion) to a direct matomo tag. I can't remember the specifics anymore, did we decide not to use the tag manager container anymore to update the tag?

stevepiercy:
@fredvd if I was part of that decision, I don't recall. The best I can suggest is to do a search in the Plone organization, and maybe that rings a bell.

https://github.com/search?q=org%3Aplone+matomo&type=issues

fredvd:
Found the cause of the confusion here.  We don't remember, because our matomo is configured to create a new container automatically when a site is created. So we have been using mtm for some years now for training sites, using the default container generated, but you have to go to tag manager first and  pick up  the tag container code from there.

fredvd (approved):
updated to use auto generated tag container for the 2024.training.plone.org

